PR #11, opened 2018-02-15
Add pubkey generation support

FrenchBen:
Although this may not be very common, I often find myself having to feed a private key to Terraform, but then needing its public key for provider specific deployments.

This adds an equivalent public key RSA generator to the current private key, and moves the `publicKey` function to the public key file.

The input is a private key string, or file.

Checking acceptance tests:
```
$ make testacc
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v  -timeout 120m
?   	github.com/terraform-providers/terraform-provider-tls	[no test files]
=== RUN   TestProvider
--- PASS: TestProvider (0.00s)
=== RUN   TestCertRequest
--- PASS: TestCertRequest (0.04s)
=== RUN   TestLocallySignedCert
--- PASS: TestLocallySignedCert (0.02s)
=== RUN   TestPrivateKeyRSA
--- PASS: TestPrivateKeyRSA (3.21s)
=== RUN   TestPrivateKeyECDSA
--- PASS: TestPrivateKeyECDSA (0.05s)
=== RUN   TestPublicKeyRSA
--- PASS: TestPublicKeyRSA (4.69s)
=== RUN   TestSelfSignedCert
--- PASS: TestSelfSignedCert (0.04s)
PASS
ok  	github.com/terraform-providers/terraform-provider-tls/tls	8.077s

### CODE REVIEW

**Comment by FrenchBen** [comment]
There's a good amount of duplicate code with the `tls_private_key` module, perhaps we can simply add support to the generation to have a private key string submitted, in which case it's not generated?

**Comment by vancluever** [comment]
Hey @FrenchBen, this might also answer your second question as well, but I think this would be better as a data source. Main reason being is that a private key already needs to exist to extract the public key from, so nothing is actually being created by the resource.

More specifically answering your question about reuse - you could probably take extraction code starting [here](https://github.com/terraform-providers/terraform-provider-tls/blob/master/tls/resource_private_key.go#L133) and throw it into a helper that both `tls_private_key` (the resource) and `tls_public_key` (the data source) could use.

File: tls/data_source_public_key.go [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+	"path/filepath"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourcePublicKey() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourcePublicKeyRead,
+		Schema: map[string]*schema.Schema{
+			"private_key": &schema.Schema{
```
vancluever: Can this be changed to `private_key_pem` to bring it in line with the other resources in this provider?

tls/data_source_public_key.go
```diff
@@ -0,0 +1,75 @@
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "PEM formatted string to use as the private key",
+			},
+
+			"private_key_path": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "File path of the PEM formatted string to use as the private key",
+				ForceNew:    true,
```
vancluever: `ForceNew` is not needed in data sources. This generally controls deletion/creation workflows, both of which data sources lack.

tls/data_source_public_key.go
```diff
@@ -0,0 +1,75 @@
...
+	return &schema.Resource{
+		Read: dataSourcePublicKeyRead,
+		Schema: map[string]*schema.Schema{
+			"private_key": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "PEM formatted string to use as the private key",
+			},
+
+			"private_key_path": &schema.Schema{
```
vancluever: This attribute should conflict with `private_key_pem` and vice versa so that only one is supplied.

tls/data_source_public_key.go
```diff
@@ -0,0 +1,75 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourcePublicKeyRead(d *schema.ResourceData, meta interface{}) error {
+	// Read private key
+	bytes := []byte(d.Get("private_key").(string))
```
vancluever: You could also use [`GetOk`](https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_data.go#L91) or [`GetOkExists`](https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_data.go#L119) here, which are specifically the exact ways you check for a value defined in Terraform.

tls/data_source_public_key.go
```diff
@@ -0,0 +1,75 @@
...
+		if err != nil {
+			fmt.Printf("%v\n", err)
+			os.Exit(1)
+		}
+	}
+	// decode PEM encoding to ANS.1 PKCS1 DER
+	keyPemBlock, _ := pem.Decode(bytes)
+
+	if keyPemBlock == nil || (keyPemBlock.Type != "RSA PRIVATE KEY" && keyPemBlock.Type != "EC PRIVATE KEY") {
+		return fmt.Errorf("failed to decode PEM block containing RSA private key")
```
vancluever: The error message seems to indicate that you are only attempting to decode an RSA private key even though the key can be that or an EC private key.

tls/data_source_public_key.go
```diff
@@ -0,0 +1,75 @@
...
+
+	keyPem := string(pem.EncodeToMemory(keyPemBlock))
+
+	// Converts an RSA private key from its ASN.1 PKCS#1 DER encoded form
+	rsaKey, err := x509.ParsePKCS1PrivateKey(keyPemBlock.Bytes)
+	if nil != err {
+		return fmt.Errorf("error converting key to rsa %s", err)
+	}
+
+	d.Set("private_key_pem", keyPem)
```
vancluever: Looks like you have the field here named `private_key_pem` even though the field is just `private_key` in the schema. I'm recommending the schema be fixed, of course. :)

tls/util.go
```diff
@@ -74,3 +75,29 @@ func parseCertificateRequest(d *schema.ResourceData, pemKey string) (*x509.Certi
 
 	return certReq, nil
 }
+
+func parsePublicKey(d *schema.ResourceData, rsaKey interface{}) error {
```
vancluever: _Minor_ can we change this to `readPublicKey` to imply that it's doing a bit more than just parsing the public key?

tls/data_source_public_key_test.go
```diff
@@ -0,0 +1,42 @@
...
+					resource.TestCheckResourceAttr("data.tls_public_key.test", "public_key_pem", strings.TrimSpace(expectedPublic)+"\n"),
+					resource.TestCheckResourceAttr("data.tls_public_key.test", "public_key_openssh", strings.TrimSpace(expectedPublicSSH)+"\n"),
+				),
+			},
+		},
+	})
+}
+
+const testAccDataSourcePublicKeyConfig = `
+data "tls_public_key" "test" {
```
vancluever: PS: You could probably make this test self-hosting but using the `tls_private_key` resource to generate a key. Then you could use [`TestCheckResourceAttrPair`](https://github.com/hashicorp/terraform/blob/master/helper/resource/testing.go#L1016) to just compare the two values.

tls/data_source_public_key.go
FrenchBen: Great point! It'll make things a lot easier.

tls/data_source_public_key_test.go
FrenchBen: I'll add it as another test. I like the fact that we're taking a private key without it's matching Public Key and confirming that its output is what we expected.

vancluever: @FrenchBen fair enough, and I actually didn't notice the `testPrivateKey` fixture being used in other tests. Will wait for the second test!

FrenchBen:
@vancluever PTAL at the latest changes :)

vancluever:
@FrenchBen sorry for leaving this so long - I'll have a look sometime within the next couple of days for sure. Were you going to work on adding the second test [as mentioned](https://github.com/terraform-providers/terraform-provider-tls/pull/11#discussion_r169429890)?

FrenchBen:
@vancluever already did:
https://github.com/terraform-providers/terraform-provider-tls/pull/11/files#diff-db53aa662810ae9abc7cf890db96dc69R50

The 2 tests are:
* Given a known private key, verify that the public key matches expected result.
* Generate a private key, verify that the public key matches that of the generated result.

vancluever:
Great - I didn't see that!

Barring anything serious, I think this can be merged tomorrow when I give it another once over :slightly_smiling_face:

FrenchBen:
I think it's my fault - Looks like spacing is a little off, I'll get this fixed.

tls/data_source_public_key_test.go
```diff
@@ -0,0 +1,64 @@
...
+					resource.TestCheckResourceAttr("data.tls_public_key.test", "public_key_pem", strings.TrimSpace(expectedPublic)+"\n"),
+					resource.TestCheckResourceAttr("data.tls_public_key.test", "public_key_openssh", strings.TrimSpace(expectedPublicSSH)+"\n"),
+				),
+			},
+			resource.TestStep{
+				Config: `
+					resource "tls_private_key" "test" {
+						algorithm = "RSA"
+					}
+					output "private_key_pem" {
```
vancluever: Are these outputs actually being tested anywhere? I don't see any checks for them...

Can this test be broken out into another test? As it's pretty much a completely different test case (manually supplied key versus generated one).

FrenchBen: The output can be removed indeed. Since the test was for all Public Key, I thought they belonged within the same function, but as separate steps. Is that not the right approach?

FrenchBen:
I'll take a look at why it's not working as expected - What's interesting is that the generation works as expected, but feeding the key to the data doesn't.
WORKS:
```hcl
resource "tls_private_key" "key" {
  algorithm   = "ECDSA"
  ecdsa_curve = "P384"
}
output "ecdsa_private_key" {
  value = "${tls_private_key.key.private_key_pem}"
}
```

FrenchBen:
Added the extra test - Different algo = different parser (of course 🤦‍♂️ ) - Moved to using the default private key parser in the certs to make things cleaner.

vancluever:
@FrenchBen this looks good! I've just split out the tests so that it's more clear what exactly we are testing in each step.

Otherwise, this fixed everything - smoke test works now, so I think we are good to merge!

Thank you very much for you contribution and your patience through the process!

PR #18, opened 2018-04-01
Fix #14: make changing elements in the subject force a new resource 

### LINKED ISSUE #14: Changing a tls_cert_request's subject does not create a new cert_request_pem
terraform version 0.11.4
provider version 1.1.0

will provide a repro later

SpencerBrown:
rather than failing with "doesn't support update" message.

Also, fix incorrect `make bin` command in README, should be `make build`


SpencerBrown:
@apparentlymart @vancluever pinging to see if we can get this PR reviewed/merged, I've got another one waiting on this one.... thanks

PR #19, opened 2018-04-11
Typo fix for ECDSA curve type

sjones-sot:
P251 > P521


mildwonkey:
Whoops! Thanks @sjones-sot, great catch.

PR #60, opened 2019-10-04
Update all dependencies and fix vendor directory

invidian:
`vendor` directory was out of sync with defined go modules. This PR fixes that and also updates all dependencies to latest versions.


invidian:
Any updates on this?

kmoe:
Thanks for the PR, @invidian. It's now been superseded by https://github.com/hashicorp/terraform-provider-tls/pull/77. This is through no fault of your own - we're currently improving our internal processes for maintaining this provider and we weren't able to respond to your PR before now.

Regarding updating all dependencies, we generally don't do this as a policy unless there's a reason to update, such as a critical security fix. So just fixing the vendor directory would have been sufficient.

Thanks for your patience!

PR #64, opened 2019-12-06
Moving the module name to add /v2 at the end.

lazzurs:
Doing this to comply with go module versioning. This is causing an issue
with moving the acme terraform provider over to the latest version of
this provider as it includes this provider as a go module.

This change will require consumers of this provider as a module to
change how they reference this module.

https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
-module github.com/terraform-providers/terraform-provider-tls
+module github.com/terraform-providers/terraform-provider-tls/v2
 
 require (
 	github.com/hashicorp/terraform-plugin-sdk v1.0.0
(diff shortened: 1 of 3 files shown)


lazzurs:
The test build has failed because this is a breaking change that needs merged/released before the build will work again.

martinssipenko:
Any updates on this?

The change proposed here is also a requirement for updating github provider as it too includes this module and right now is stuck with version 1.

kmoe:
Importing providers as Go modules is not supported (please see [best practice documentation](https://www.terraform.io/docs/extend/best-practices/depending-on-providers.html#do-not-import-providers-as-go-modules)), so this is not necessary.

Issue #56, opened 2019-09-30
Creating a CA certificate should set Authority Key Identifier attribute


seboudry:
Hi !

By reviewing the code, I found a little mistake when creating a a CA self-signed certificate.
Resource: `tls_self_signed_cert`

https://github.com/terraform-providers/terraform-provider-tls/blob/cb21e556b00b4d8d5fc4073b941ddd5a9c2c0395/tls/resource_certificate.go#L169
Should `AuthorityKeyId` be set instead of `SubjectKeyId` when using `is_ca_certificate`?

binlab:
The same issue I saw, when we creating a CA is `X509v3 Authority Key Identifier` extension missed. Could you comment on this behavior?

Sebor:
I have the same issue. Moreover I suspect that it causes problem for windows OS (because I get `NET::ERR_CERT_AUTHORITY_INVALID` even after importing root CA)

karelorigin:
https://github.com/hashicorp/terraform-provider-tls/pull/66 is supposed to patch this issue but hasn't been active in a while. Any updates from HashiCorp would be great.

PR #66, opened 2019-12-10
Add Authority Key Identifier to locally signed certificate

Q69K:


karelorigin:
Encountered this issue when trying to import a Terraform generated CA intermediate into Vault. It's been open for years, would be nice if we could get this merged :)

Looks like the PR author still needs to sign the CLA agreement. CC: @Q69K

irigon:
Found this issue, as some certificates stop working as we updated the OpenSSL version to over 1.1.1k.
When both CA and server CSR have the same subject, the field Key Identifier is omitted in the certificate. 
Changing the CN on the server certificate CSR or adding additional fields to the subject was enough to cause the X509v3 Authority Key Identifier field to the added to the certificate.

detro:
Hello and thanks for the contribution.

Because no example was provided to reproduce this, I have tested it and found that, for `tls_locally_signed_cert`, if the provided CA carries a value for `SubjectKeyId`, than the underlying [`x509.CreateCertificate`](https://pkg.go.dev/crypto/x509#CreateCertificate) function will indeed populate the `AuthorityKeyIdentifier` (please [see here](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/crypto/x509/x509.go;l=1520)).

Additionally, local testing where I tried to reproduce, showed that the behaviour is what the `x509` library correctly does.

This is NOT the case for `tls_self_signed_cert`, where in the current implementation the `AuthorityKeyIdentifier` is indeed not set (as reported in #56).

As I started working on this together with #56, I'll have a PR incoming that will close both.
But I just wanted to emphasize that the issue reported here lies in the provided CA to the `tls_locally_signed_cert` resource, not the resource itself.

Thank you.

PR #69, opened 2020-03-24
Add PKCS12 archive bundle to self-signed certificate resource

### LINKED ISSUE #29: Resource to generate a PKCS #12 archive file out of PKI components
It would be useful to support generating a PKCS #12 (PKCS12 or also known as PFX) archive file out of PKI components, including an optional private key, and a chain of one or more certificates.  The current set of TLS resources support generating/exporting to PEM format files, but there are many cases where a PFX archive file is preferred or required such as Windows host environments, Azure provider or Oracle OCI provider.

I recommend supporting this resource in a more general-purpose manner, such that you can give it an optional private key PEM, and one or more certificates (i.e. an array of certificate PEM clobs).  By doing so, the resource would be usable in tandem with other resources of this provider, as well as alternate providers, such as the [third-party ACME cert provider](https://github.com/vancluever/terraform-provider-acme).

An example usage might look like this:
```hcl
## Generate a self-signed cert using a private key stored on disk
resource "tls_self_signed_cert" "example" {
  key_algorithm   = "ECDSA"
  private_key_pem = "${file(\"private_key.pem\")}"

  subject {
    common_name  = "example.com"
    organization = "HCTF Examples, Inc"
  }

  validity_period_hours = 12

  allowed_uses = [
    "key_encipherment",
    "digital_signature",
    "server_auth",
  ]
}

## Produce a PKCS #12 archive of the self-signed cert (private key included)
resource "tls_pkcs12_archive" "example_pfx" {
  private_key_pem = "${file(\"private_key.pem\")}"
  certificate_pem = [
    "${tls_self_signed_cert.example.cert_pem}"
  ]
}

## Save the PKCS #12 archive to an S3 bucket
resource "aws_s3_bucket_object" "example_pfx_s3obj" {
    bucket = "${aws_s3_bucket.example_s3_bucket.id}"
    key    = "pki/example.pfx"

    content = <<EOF
${tls_pkcs12_archive.example_pfx.archive_blob}
EOF
}

```

### Terraform Version
Terraform v0.11.3

### Affected Resource(s)
provider.tls v1.1.0

vadimkuznetsov:
Resolves #29

### DIFF
--- a/tls/resource_certificate.go
+++ b/tls/resource_certificate.go
@@ -137,11 +137,25 @@ func resourceCertificateCommonSchema() map[string]*schema.Schema {
 			Description: "If true, the generated certificate will include a subject key identifier.",
 			ForceNew:    true,
 		},
+
+		"certificate_p12": {
+			Type:      schema.TypeString,
+			Computed:  true,
+			Sensitive: true,
+		},
+
+		"certificate_p12_password": {
+			Type:      schema.TypeString,
+			Optional:  true,
+			Default:   "",
+			Sensitive: true,
+		},
 	}
 }
 
 func createCertificate(d *schema.ResourceData, template, parent *x509.Certificate, pub crypto.PublicKey, priv interface{}) error {
 	var err error
+	var caCerts []*x509.Certificate
 
 	template.NotBefore = now()
 	template.NotAfter = template.NotBefore.Add(time.Duration(d.Get("validity_period_hours").(int)) * time.Hour)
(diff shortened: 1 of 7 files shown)


tiwood:
@appilon, any change is gets reviewed in the next days?
We would love to have this functionality.

tls/resource_certificate.go
```diff
@@ -185,6 +199,26 @@ func createCertificate(d *schema.ResourceData, template, parent *x509.Certificat
 	}
 	certPem := string(pem.EncodeToMemory(&pem.Block{Type: pemCertType, Bytes: certBytes}))
 
+	private_key, err := parsePrivateKey(d, "private_key_pem", "key_algorithm")
+	// Set PKCS12 data. This is only set if there is a private key present.
+	if err != nil {
```
tiwood, line 204: I think this should be inverted.
You want to go into the `if` block if no error is given.

tls/resource_certificate.go
```diff
@@ -185,6 +199,26 @@ func createCertificate(d *schema.ResourceData, template, parent *x509.Certificat
...
+		cert, err := x509.ParseCertificate(certBytes)
+		if err != nil {
+			return err
+		}
+		caCerts = append(caCerts, parent)
```
tiwood, line 209: I tried this out and this gave me troubles in the `PFX` validation.
If the parent (which in this case is the same as the cert as `cert`) is provided the p12/pkcs12 is available but testing it with `openssl pkcs12 -info` gave me errors:
```
CRYPTO_internal:header too long
CRYPTO_internal:bad object header
CRYPTO_internal:nested asn1 error
CRYPTO_internal:decode error
```

Leaving `caCerts` empty, gave me a valid p12/pkcs12.

tls/resource_certificate.go
vadimkuznetsov: fixed

tls/resource_certificate.go
vadimkuznetsov: removed

vadimkuznetsov:
> I've tested this PR on my side and had to make 2 changes to get it running.
Thanks for testing. :)

vadimkuznetsov:
Any ideas how to fix ci test?

elliotchaim:
Hi guys any update on this?

devlincashman:
This would be a massive feature for us as we use Azure and they almost universally expect certificates in the pfx format. Currently we do lots and lots of workarounds with local-exec and external provider to manage this. Would really like to see this feature added. PFX support is crucial as it is the main type of certificate in the Microsoft ecosystem (for better or worse).

detro:
Superseded by #205.

PR #73, opened 2020-04-05
Implemented tls_x509_crl resource

### LINKED ISSUE #20: Certificate revocation list
Use case is the following:
If you have a server which requires a valid certificate to log in, you would also want to keep the CRL of the revoked certificates in order to prevent the deleted users get authorized.
When you iterate over a list with users, you need to create new certificates for new users and revoke the certificates for the deleted users, which have disappeared from the list. 

I'm not sure whether the current architecture of terraform states allows us to keep the history of changes, but we need to think about generating the CRL somehow. In theory, we may use some kind of index file to keep the history of certificates.

fllaca:
Hi, I added here a first version of a x509 CRL resource, resolving #20 . The way of using it is using a field `certs_to_revoke`, that contains the certificates (in PEM format) that will be revoked.

Resolves #20


tls/resource_cert_revocation_list.go
```diff
@@ -0,0 +1,193 @@
...
+func resourceCertRevocationList() *schema.Resource {
+	s := map[string]*schema.Schema{
+		"certs_to_revoke": &schema.Schema{
+			Type: schema.TypeList,
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+			Required:    true,
+			Description: "PEM-encoded certificates to be revoked",
+			ForceNew:    true,
```
invidian, line 23: While technically correct, it would be better, if we would update the list of certificates in the CRL, rather than create the new one every time, as then the expiry date will be updated too for example.

tls/resource_cert_revocation_list.go
```diff
@@ -0,0 +1,193 @@
...
+			Sensitive:   true,
+			StateFunc: func(v interface{}) string {
+				return hashForState(v.(string))
+			},
+		},
+		"ca_key_algorithm": &schema.Schema{
+			Type:        schema.TypeString,
+			Required:    true,
+			Description: "Name of the algorithm used to generate the certificate's private key",
```
invidian, line 79: This description doesn't seem right.

tls/resource_cert_revocation_list.go
```diff
@@ -0,0 +1,193 @@
...
+func DeleteCRL(d *schema.ResourceData, meta interface{}) error {
+	d.SetId("")
+	return nil
+}
+
+func ReadCRL(d *schema.ResourceData, meta interface{}) error {
+	return nil
+}
+
+func UpdateCRL(d *schema.ResourceData, meta interface{}) error {
```
invidian, line 157: I think this function can be removed for now, if we don't support updates.

tls/resource_cert_revocation_list.go
```diff
@@ -0,0 +1,193 @@
...
+		endTime = endTime.Add(earlyRenewalPeriod)
+
+		currentTime := now()
+		timeToRenewal := endTime.Sub(currentTime)
+		if timeToRenewal <= 0 {
+			readyForRenewal = true
+		}
+	}
+
+	if readyForRenewal {
```
invidian: I would move this block of code to separate function, let's call it `markForRenewal()`, so then at the beginning of this function, you could do the following:
```go
	err := endTime.UnmarshalText([]byte(endTimeStr))
	if err != nil {
          return markForRenewal(d)
        }
```

This way, you can avoid using `else`, which generally should be avoided in Go (e.g. https://medium.com/@jamousgeorges/dont-use-else-66ee163deac3).

`markForRenewal()` would be something like:
```go
func markForRenewal(d *schema.ResourceDiff) error {
  err := d.SetNew("ready_for_renewal", true)
  if err != nil {
    return err
  }

  return d.ForceNew("ready_for_renewal")
}
```

tls/resource_cert_revocation_list.go
```diff
@@ -0,0 +1,193 @@
...
+
+func CreateCRL(d *schema.ResourceData, meta interface{}) error {
+	certsToRevoke := make([]pkix.RevokedCertificate, 0)
+	if value := d.Get("certs_to_revoke").([]interface{}); len(value) > 0 {
+		for _, vi := range value {
+			block, err := decodePEMFromBytes([]byte(vi.(string)), "")
+			if err != nil {
+				return err
+			}
+			certificate, err := x509.ParseCertificate(block.Bytes)
```
invidian, line 102: I wonder if `parseCertificate` could be somehow refactored, so it can be used here and in the tests. I guess PEM decoding + parsing x509 could be always done together.

tls/resource_cert_revocation_list.go
```diff
@@ -0,0 +1,193 @@
...
+		return err
+	}
+
+	crlBytes, err := caCert.CreateCRL(rand.Reader, caKey, certsToRevoke, notBefore, notAfter)
+	if err != nil {
+		return err
+	}
+
+	crlPem := string(pem.EncodeToMemory(&pem.Block{Type: "X509 CRL", Bytes: crlBytes}))
+	if err != nil {
```
invidian, line 137: This error check is not needed I think.

tls/resource_cert_revocation_list_test.go
```diff
@@ -0,0 +1,89 @@
...
+			{
+				Config: certRevocationListConfig(1, 0),
+				Check: func(s *terraform.State) error {
+					gotUntyped := s.RootModule().Outputs["crl_pem"].Value
+					got, ok := gotUntyped.(string)
+					if !ok {
+						return fmt.Errorf("output for \"crl_pem\" is not a string")
+					}
+					if !strings.HasPrefix(got, "-----BEGIN X509 CRL----") {
+						return fmt.Errorf("key is missing cert PEM preamble")
```
invidian, line 28:
```suggestion
						return fmt.Errorf("key is missing CRL PEM preamble")
```

tls/resource_cert_revocation_list_test.go
```diff
@@ -0,0 +1,89 @@
...
+					}
+					block, _ := pem.Decode([]byte(got))
+					crl, err := x509.ParseCRL(block.Bytes)
+					if err != nil {
+						return fmt.Errorf("error parsing crl: %s", err)
+					}
+
+					// Verify testCertificate is in the CRL serial number list
```
invidian, line 36:
```suggestion
					// Verify testCertificate is in the CRL serial number list.
```

tls/resource_cert_revocation_list_test.go
```diff
@@ -0,0 +1,89 @@
...
+					testCertBlock, _ := pem.Decode([]byte(testCertificate))
+					testCert, err := x509.ParseCertificate(testCertBlock.Bytes)
+					if err != nil {
+						return fmt.Errorf("error parsing test cert: %s", err)
+					}
+					if testCert.SerialNumber.Cmp(crl.TBSCertList.RevokedCertificates[0].SerialNumber) != 0 {
+						return fmt.Errorf("revoked certificate serial number doesn't match")
+					}
+
+					// Verify CRL signature with CA
```
invidian, line 46:
```suggestion
					// Verify CRL signature with CA.
```

tls/resource_cert_revocation_list_test.go
```diff
@@ -0,0 +1,89 @@
...
+				},
+			},
+		},
+	})
+}
+
+func certRevocationListConfig(validity uint32, earlyRenewal uint32) string {
+	return fmt.Sprintf(`
+	                locals {
+						cert_to_revoke = <<EOT
```
invidian, line 220: I think indentation here could be improved. Perhaps tabs and spaces are mixed?

tls/resource_cert_revocation_list.go
```diff
@@ -0,0 +1,193 @@
...
+				return err
+			}
+			certsToRevoke = append(certsToRevoke, pkix.RevokedCertificate{
+				SerialNumber: certificate.SerialNumber,
+			})
+		}
+	}
+	caKey, err := parsePrivateKey(d, "ca_private_key_pem", "ca_key_algorithm")
+	if err != nil {
+		return err
```
invidian, line 113: I think it would be useful to annotate those errors using `fmt.Errorf`, so the error message is more clear for the user:
```suggestion
		return fmt.Errorf("failed to parse %q field: %w", err)
```

Also for other `return err`, which faces user.

tls/provider_test.go
```diff
@@ -96,6 +96,26 @@ UI+cIbsQJcr9i0SADwhqs7XYIcEz9jDqrdezenaXdGtXYg==
 -----END CERTIFICATE-----
 `
 
+var testCertificate = `
```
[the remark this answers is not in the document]
fllaca: I just thought this would be closer to a real usage, should I change the test to use caCertificate and remove this?

tls/resource_cert_revocation_list.go
fllaca: Yeah... but I'm quite unsure on how to handle that lifecycle 🤔 , for instance the Id of the resource is calculated as a hash of the CRL PEM content (in the same way we do for locally signed certificates and other resources in this provider). When a CRL is "updated" actually what happens is that a new version of it is created, basically a new CRL itself. I was checking https://tools.ietf.org/html/rfc5280#section-5.1 trying to find a field that we can use to identify a CRL independently of the version, but I couldn't find any. But I can miss something, do you have any idea around this?

tls/resource_cert_revocation_list.go
fllaca: It was kept because of the `early_renewal_hours` field. If we make the resource non-updatable that field needs to be flagged with `ForceNew: true`. Does it make sense?

fllaca:
Hi @invidian , thanks a lot for hour review! I addressed most of your comments, please take a look again when you have a minute. For some other comments I had some doubts, responded with some thoughts about them.

tls/resource_cert_revocation_list.go
fllaca: I can do that very easily, I just followed the same error handling in other parts of the code that also do this private key parsing, for consistency with the rest of the code. Should I change it here anyway? I can also change it in the other parts a key or certificate is parsed.

tls/resource_cert_revocation_list_test.go
fllaca: should be fixed now

tls/resource_cert_revocation_list.go
fllaca: Done, love this suggestion! it reduces the cognitive complexity making it way easier to understand, thanks! Thanks also for the article reference, interesting! :)

tls/resource_cert_revocation_list_test.go
```diff
@@ -0,0 +1,89 @@
+package tls
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+
+	//"errors"
```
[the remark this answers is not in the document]
fllaca: removed!

tls/resource_cert_revocation_list.go
```diff
@@ -0,0 +1,193 @@
...
+		Read:          ReadCRL,
+		Update:        UpdateCRL,
+		CustomizeDiff: CustomizeCertificateDiff,
+		Schema:        s,
+	}
+}
+
+func CreateCRL(d *schema.ResourceData, meta interface{}) error {
+	certsToRevoke := make([]pkix.RevokedCertificate, 0)
+	if value := d.Get("certs_to_revoke").([]interface{}); len(value) > 0 {
```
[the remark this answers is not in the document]
fllaca: Done, also added a test for the case this field is empty

tls/resource_cert_revocation_list.go
fllaca: Done!

tls/provider_test.go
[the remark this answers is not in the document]
invidian: I was just wondering if there is some specific reason why separate certificate. I think it should be fine to re-use CA certificate, as for CA certificates, you can also use the CRLs, the same as for "end" certificates.

In my opinion, the maintainers have a deciding vote in such cases, but I would personally try to reuse the assets, which are already in place.

tls/resource_cert_revocation_list.go
fllaca: right, removed

tls/resource_cert_revocation_list.go
invidian: Hm, I'm not sure if I understand the problem. If you set the ID only in `Create` function and then don't touch it in `Update`, then what's the problem? I don't think there is a need to change the ID.

Aside from that, I think this is completely fine for the initial implementation. It can always be improved later :+1:

tls/resource_cert_revocation_list.go
invidian: In my opinion, ideally all the errors should be wrapped, as far as I know it's considered a good practice (https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-
gracefully) as it allows to quickly identify what exactly broke and provide better context for the user.

The comment I left was just an example on one line (pointing out every `return err` would be a lot of work :smile:), so for any new code, I'd try to follow this. I just wouldn't change the existing codebase, as this should be done in separate context (PR/issue). I hope that make sense.

tls/provider_test.go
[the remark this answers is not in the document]
fllaca: Agree, I juste removed the `testCert` variable and reused testCACert for this test

tls/resource_cert_revocation_list.go
fllaca: Agree, done! wrapped the errors of the new code in this PR

tls/resource_cert_revocation_list_test.go
```diff
@@ -0,0 +1,298 @@
...
+                        ca_cert_pem = <<EOT
+%s
+EOT
+                        ca_key_algorithm = "RSA"
+                        ca_private_key_pem = <<EOT
+%s
+EOT
+                    }
+                    output "crl_pem" {
+                        value = "${tls_x509_crl.test.crl_pem}"
```
invidian, line 295: Ditto `${}`.

fllaca:
Hi @invidian , sorry for the delayed response, I was really busy the last two weeks.

Thank you so much for the thoughtful feedback! Yeah... I see the points in your argumentation: as you say, with the current approach in this PR if you revoke a certificate and add it to the CRL you need to keep it forever in the terraform code and state (either hardcoded PEM or keeping its certificate resource definition). If you taint a certificate, the old version of it is not going to be in the list of serial numbers of the newly regenerated CRL. 

But on the other hand, if you taint it, from the crypto point of view it is going to be actually a completely new certificate different from the previous one you had in tf state. I'm not that sure that the Terraform resource should handle this usecase internally by maintaining an "history" of revoked certificates in the state that don't appear in the resource definition in the terraform code. Normally, what I expect from terraform is that the code, the state, and the actual real resources are three views of the same thing, and Terraform's goal is to reconcile all three. Anyway I can be wrong here, I don't know if there's any other existing provider/resource that does something similar 🤔 , it could serve as inspiration to implement this as well.

These are just some thoughts, but I don't have a strong opinion... I agree with you, it would be good having some more input feedback on this. In any case I will give it a try to implement a version that keeps the history of revoked certs, maybe in another branch, at least just to see how it looks.

mhmdio:
any update on this @fllaca ?

swade1987:
any update on this?

plsanchezfaure:
any update on this?

sergio-toro:
Hey any update on this PR? would be great to manage revocation lists with this module 🙏🏽

trebidav:
Would love to see this implemented! It's needed for years already!

babicamir:
Any updates!?
Security concerns are growing!?
Thanks

circa10a:
Still growing!

PR #76, opened 2020-04-10
Implement support for x509 extra extensions

AbdulRahmanAlHamali:
This PR provides the ability to add arbitrary extensions to certificates and certificate requests, by providing their OID, whether they are critical or not, and their value.

The value of the extension is serialized into Abstract Syntax Notation One (ASN.1)

The PR only implements a few of the possible types for value, but makes it very easy to add more types upon demand.

Please let me know what you think, looking forward to the review.

Regards,


tls/provider.go
```diff
@@ -121,3 +126,109 @@ var nameSchema *schema.Resource = &schema.Resource{
...
+			Description: "Whether the extension should be treated as critical",
+			Optional:    true,
+			Default:     false,
+			ForceNew:    true,
+		},
+		"integer_value": {
+			Type:        schema.TypeInt,
+			Description: "Fill this field if the extension value should be encoded as an ASN.1 INTEGER",
+			Optional:    true,
+			ForceNew:    true,
```
julienduchesne, line 206: You can use the `ConflictsWith` attribute to make sure that only one is set:
```
ConflictsWith: []string{"boolean_value", "printable_string_value", "utf8_string_value"},
```

tls/provider.go
```diff
@@ -121,3 +126,109 @@ var nameSchema *schema.Resource = &schema.Resource{
...
+		},
+		"utf8_string_value": {
+			Type:        schema.TypeString,
+			Description: "Fill this field if the extension value should be encoded as an ASN.1 UTF8String",
+			Optional:    true,
+			ForceNew:    true,
+		},
+		"type": {
```
julienduchesne, line 226: If you have all the `ConflictsValue` set right, you no longer need this since only one of the values above will be set. The `GetOk` method should give you the `result` and `ok` (whether or not the value has been set by the user)

tls/resource_cert_request_test.go
```diff
@@ -43,7 +45,25 @@ func TestCertRequest(t *testing.T) {
                         uris = [
                             "spiffe://example-trust-domain/workload",
                             "spiffe://example-trust-domain/workload2",
-                        ]
+						]
```
julienduchesne, line 48: formatting seems weird here

tls/provider.go
AbdulRahmanAlHamali: Awesome! Didn't know about this

tls/provider.go
AbdulRahmanAlHamali: This won't work unfortunately. `ConflictsWith` does not work on nested resources:
https://github.com/hashicorp/terraform-plugin-sdk/issues/71

And I can't call `GetOk` on a nested resource, because it is returned to me as a map instead of as `ResourceData`

Karthikeyanraman94:
HI Team, can we have this merged, we literally need this feature for one of our client certificate

Karthikeyanraman94:
@julienduchesne : could you please this feature go live

Karthikeyanraman94:
@julienduchesne : Could you please help this feature go live... We actually require OID feature for few of our client certs.. And this will be more helpfull

PR #88, opened 2020-11-26
Fix package to follow correct Go version guidelines

vancluever:
The current package name
(github.com/terraform-providers/terraform-provider-tls) lacks a version
path and as such cannot be imported into other projects.

I currently depend on this provider in
github.com/vancluever/terraform-provider-acme, and this is effectively
keeping me from modernizing the provider due to a lack of ability to
upgrade testing dependencies.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/terraform-providers/terraform-provider-tls
+module github.com/terraform-providers/terraform-provider-tls/v3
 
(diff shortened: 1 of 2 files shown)


vancluever:
Closing this. Looks like the package was moved internally, so I will need to determine other means to use external provider dependencies in my test now. I'll look at the docs to see what the correct pattern is.

PR #130, opened 2021-09-08
Add sha256 fingerprint alongside sha1

benohara:
Generate SHA256 fingerprint and also SHA1/256 fingerprints
in RFC4716 HEX format.


acolombier:
Thanks for your work! Any chance to get this in soon? I'm also looking for that feature

Issue #150, opened 2022-02-17
Add support for ED25519 key algorithm


detro:
## Background

Support for the [ED25519](https://ed25519.cr.yp.to/) key algorithm has been a community-requested feature for this provider for a while. The first issue we could track down was filed in May 2018 ([#26](https://github.com/hashicorp/terraform-provider-tls/issues/26)). Since then, multiple contributors attempted to bridge this gap ([#32](https://github.com/hashicorp/terraform-provider-tls/pull/32) in Nov 2018, [#59](https://github.com/hashicorp/terraform-provider-tls/pull/59) in Oct 2019, [#85](https://github.com/hashicorp/terraform-provider-tls/pull/85) in Oct 2020).

We want to take this opportunity to thank all the contributors (@jbboehr, @kaidence, @invidian, @nikolay) for the PR they submitted, and also to apologize for the long delay in getting this set of functionalities added to this provider. HashiCorp has a policy to prioritize stability and bug-fixes over any new feature for our “utility providers”; but given the demand for this feature, we decided to look into this and see how we could fit ED25519 inside the feature set.

## Proposal

To serve the community at our best, we want to provide a well-rounded implementation that adds support for ED25519. Specifically, we want it to match the same functionality coverage as per the currently-supported [RSA](https://en.wikipedia.org/wiki/RSA_(cryptosystem)) and [ECDSA](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm).

The implementation will include changes to the following entities, so that they accept `ED25519` as input in their `algorithm / key_algorithm` configuration fields:

**Resources:**

* [x] `tls_private_key` (#151)
  * Additional field: `private_key_openssh`
* [x] `tls_cert_request` (#173)
* [x] `tls_locally_signed_cert` (#173)
* [x] `tls_self_signed_cert` (#173)

**Data sources:**
* [x] `tls_public_key` (#160)

As work proceeds, PRs will be linked back to this issue for full context.

## References

Addresses: [#26](https://github.com/hashicorp/terraform-provider-tls/issues/26)
Supersedes: [#32](https://github.com/hashicorp/terraform-provider-tls/pull/32)
Supersedes: [#59](https://github.com/hashicorp/terraform-provider-tls/pull/59)
Supersedes: [#85](https://github.com/hashicorp/terraform-provider-tls/pull/85)

tazjin:
[...]

PR #59, opened 2019-10-03
r/private_key: Add support for ed25519 algorithm

invidian:
This PR adds support for generating private keys using `ed25519` algorithm. 

NOTE: I wasn't sure what PEM header to use for this certificate. Since my usecase is to generate SSH keys, I used `OPENSSH PRIVATE KEY` for now and also the output of PEM private key is currently in SSH-compatible format. Please guide me how this should be handled, as I think this output is valuable. Maybe we could have new computed property `private_key_openssh` with properly formatted all keys for SSH? BTW this problem does not occur with RSA and ECDSA keys, since `sshd` accepts those keys with output provided by this resource.

Also included some basic unit tests, please let me know if more conditions should be added.

Refs #26

PR #151, opened 2022-02-19
`tls_private_key` resource: adding support for ED25519 key algorithm

detro:
Context: #150 

This PR starts from the work of #59 (thank you!) and moves towards the use of upcoming features of `x/crypto/ssh` ([link](https://go-review.googlesource.com/c/crypto/+/218620/)).

A configuration like this will now be possible:

```hcl
resource "tls_private_key" "test" {
    algorithm = "ED25519"
}

output "private_key_pem" {
    value = "${tls_private_key.test.private_key_pem}"
    sensitive = true
}

output "private_key_openssh" {                                      #< new
    value = "${tls_private_key.test.private_key_openssh}"
    sensitive = true
}

output "public_key_pem" {
    value = "${tls_private_key.test.public_key_pem}"
}

output "public_key_openssh" {
    value = "${tls_private_key.test.public_key_openssh}"
}

output "public_key_fingerprint_md5" {
    value = "${tls_private_key.test.public_key_fingerprint_md5}"
}

output "public_key_fingerprint_sha256" {.                           #< new
    value = "${tls_private_key.test.public_key_fingerprint_sha256}"
}
```

Subtasks in this PR:

* [x] New `private_key_openssh` for all key algorithm that we can generate OpenSSH PEM format for (i.e. this excludes ECDSA P-224, because of SSH limitations)
* [x] New `public_key_fingerprint_sha256` for all key algorithms that we can generate OpenSSH PEM format for (same as above)
* [x] Extend test-set to cover the new fields
* [x] Update `tls_private_key` documentation
* [x] Extend testing matrix to include all current minor versions of Terraform

**PLEASE NOTE**: This PR does NOT cover everything we want to address in #150 but focuses around `tls_private_key`. This means that we will not be cutting a release until all the items in #150 are addressed.


go.mod
```diff
@@ -4,5 +4,6 @@ go 1.16
 
 require (
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.10.1
+	github.com/stretchr/testify v1.7.0
```
kmoe, line 7: `testify` probably not needed here - makes tests inconsistent (and makes it too easy to write tests without descriptive errors on failure).

internal/openssh/lib_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"crypto/rand"
+	"crypto/rsa"
+	"encoding/pem"
+	"golang.org/x/crypto/ssh"
+	"testing"
+
+	testifyAssert "github.com/stretchr/testify/assert"
+)
+
+func TestOpenSSHFormat_MarshalAndUnmarshall_RSA(t *testing.T) {
```
kmoe, line 16:
```suggestion
func TestOpenSSHFormat_MarshalAndUnmarshal_RSA(t *testing.T) {
```

internal/openssh/lib_test.go
```diff
@@ -0,0 +1,79 @@
...
+	rsaParsed, ok := rawPrivateKey.(*rsa.PrivateKey)
+	assert.True(ok)
+
+	// Confirm RSA is valid
+	assert.NoError(rsaParsed.Validate())
+	// Confirm it matches the original key by comparing the public ones
+	assert.True(rsaParsed.Equal(rsaOrig))
+}
+
+func TestOpenSSHFormat_MarshalAndUnmarshall_ECDSA(t *testing.T) {
```
kmoe, line 39:
```suggestion
func TestOpenSSHFormat_MarshalAndUnmarshal_ECDSA(t *testing.T) {
```

internal/openssh/lib_test.go
```diff
@@ -0,0 +1,79 @@
...
+	// Parse it back into an ECDSA private key
+	rawPrivateKey, err := ssh.ParseRawPrivateKey(pemOpenSSHPrvKeyBytes)
+	ecdsaParsed, ok := rawPrivateKey.(*ecdsa.PrivateKey)
+	assert.True(ok)
+
+	// Confirm it matches the original key by comparing the public ones
+	assert.True(ecdsaParsed.Equal(ecdsaOrig))
+}
+
+func TestOpenSSHFormat_MarshalAndUnmarshall_ED25519(t *testing.T) {
```
kmoe, line 60:
```suggestion
func TestOpenSSHFormat_MarshalAndUnmarshal_ED25519(t *testing.T) {
```

internal/provider/resource_private_key.go
```diff
@@ -71,49 +90,71 @@ func resourcePrivateKey() *schema.Resource {
...
+				Type:        schema.TypeString,
+				Description: "Public key data in OpenSSH-compatible PEM format",
+				Computed:    true,
 			},
 
 			"public_key_fingerprint_md5": {
-				Type:     schema.TypeString,
-				Computed: true,
+				Type:        schema.TypeString,
+				Description: "Fingerprint of the public key as MD5 hashing",
```
kmoe, line 126:
```suggestion
				Description: "MD5 hash of the public key",
```

internal/provider/resource_private_key.go
```diff
@@ -71,49 +90,71 @@ func resourcePrivateKey() *schema.Resource {
...
+				Computed:    true,
+			},
+
+			"public_key_fingerprint_sha256": {
+				Type:        schema.TypeString,
+				Description: "Fingerprint of the public key as SHA256 hashing",
```
kmoe, line 132:
```suggestion
				Description: "SHA256 hash of the public key",
```

internal/provider/resource_private_key.go
kmoe: There's a longer (Markdown) description for these attributes in the docs Attributes Reference section - might be better to copy this into the schema so it is not later obliterated by documentation generation.

internal/provider/resource_private_key.go
```diff
@@ -71,49 +90,71 @@ func resourcePrivateKey() *schema.Resource {
...
 	}
 
-	key, err := keyFunc(d)
+	// Generate the new Key
+	key, err := keyGen(d)
 	if err != nil {
 		return err
 	}
 
+	// Marshall the Key in PEM block
```
kmoe, line 155:
```suggestion
	// Marshal the Key in PEM block
```

internal/provider/resource_private_key.go
```diff
@@ -125,26 +166,50 @@ func CreatePrivateKey(d *schema.ResourceData, meta interface{}) error {
...
+			Bytes: keyBytes,
+		}
 	default:
 		return fmt.Errorf("unsupported private key type")
 	}
-	keyPem := string(pem.EncodeToMemory(keyPemBlock))
+	d.Set("private_key_pem", string(pem.EncodeToMemory(keyPemBlock)))
 
-	d.Set("private_key_pem", keyPem)
+	// Marshall the Key in OpenSSH PEM block, if enabled
```
kmoe, line 194:
```suggestion
	// Marshal the Key in OpenSSH PEM block, if enabled
```

internal/provider/resource_private_key_test.go
```diff
@@ -20,7 +20,12 @@ func TestPrivateKeyRSA(t *testing.T) {
                     }
                     output "private_key_pem" {
                         value = "${tls_private_key.test.private_key_pem}"
+						sensitive = true
                     }
+					output "private_key_openssh" {
```
kmoe, line 25: Tabs to spaces

internal/provider/resource_private_key_test.go
```diff
@@ -115,7 +147,12 @@ func TestPrivateKeyECDSA(t *testing.T) {
...
```
kmoe, line 152: Tabs to spaces

internal/provider/util.go
```diff
@@ -9,6 +9,15 @@ import (
 	"golang.org/x/crypto/ssh"
 )
 
+// Algorithm represents the type of private key algorithm currently supported
+type Algorithm string
```
kmoe, line 13: Opt: Could use `iota` enum pattern instead.

website/docs/r/private_key.html.md
```diff
@@ -24,43 +23,74 @@ state and does not create any external managed resources.
...
-* `ecdsa_curve` - (Optional) When `algorithm` is "ECDSA", the name of the elliptic
-curve to use. May be any one of "P224", "P256", "P384" or "P521", with "P224" as the
-default.
+~> **Limitations when using `ECDSA` with `P224` curve**
+This provider exposes attributes that can be used by [OpenSSH](https://www.openssh.com/): 
+`.public_key_openssh`, `.private_key_openssh`, `.public_key_fingerprint_md5` and `.public_key_fingerprint_sha256`.
+The SSH ECC Algorithm Integration ([RFC 5656](https://datatracker.ietf.org/doc/html/rfc5656))
+restricts support for elliptic curves to "nistp256", "nistp384" and "nistp521".
+This means that, even thought `ECDSA` and `P224` are supported here, the OpenSSH-specific fields will
+be left empty.
```
kmoe, line 63: Put most important information at beginning of warning paragraph.

```suggestion
~> **Limitations when using `ECDSA` with `P224` curve**
When using `ECDSA` with `P224`, `.public_key_openssh`, `.private_key_openssh`, `.public_key_fingerprint_md5` and `.public_key_fingerprint_sha256` will have a value of `""`.
The SSH ECC Algorithm Integration ([RFC 5656](https://datatracker.ietf.org/doc/html/rfc5656))
restricts support for elliptic curves to "nistp256", "nistp384" and "nistp521", so these [OpenSSH](https://www.openssh.com/)-specific attributes will be left blank.
```

internal/openssh/lib_test.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+func TestOpenSSHFormat_MarshalAndUnmarshall_RSA(t *testing.T) {
+	assert := testifyAssert.New(t)
+
+	// Given an RSA private key
+	rsaOrig, err := rsa.GenerateKey(rand.Reader, 4096)
+	assert.NoError(err)
+
+	// Marshall it to OpenSSH PEM format
```
kmoe, line 23:
```suggestion
	// Marshal it to OpenSSH PEM format
```

internal/openssh/lib_test.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func TestOpenSSHFormat_MarshalAndUnmarshall_ECDSA(t *testing.T) {
+	assert := testifyAssert.New(t)
+
+	// Given an ECDSA private key
+	ecdsaOrig, err := ecdsa.GenerateKey(elliptic.P521(), rand.Reader)
+	assert.NoError(err)
+
+	// Marshall it to OpenSSH PEM format
```
kmoe, line 46:
```suggestion
	// Marshal it to OpenSSH PEM format
```

internal/openssh/lib_test.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func TestOpenSSHFormat_MarshalAndUnmarshall_ED25519(t *testing.T) {
+	assert := testifyAssert.New(t)
+
+	// Given an ED25519 private key
+	_, ed25519Orig, err := ed25519.GenerateKey(rand.Reader)
+	assert.NoError(err)
+
+	// Marshall it to OpenSSH PEM format
```
kmoe, line 67:
```suggestion
	// Marshal it to OpenSSH PEM format
```

kmoe (approved):
Adding comments from verbal review. Excellent work and LGTM in general with minor changes.

internal/provider/resource_private_key.go
detro: I gave a quick read to https://datatracker.ietf.org/doc/html/rfc4716#section-4, and it's kind of "reductive" to just say "MD5 hash of the public key": it's more like that you can generate fingerprinting of a key by following an algorithm. And there is the one described in RFC4716 (that does an MD5 hash of the public key), as well as new versions like the one documented in these release notes https://www.openssh.com/txt/release-6.8.

So I'm going with:
```
Fingerprint of the public key data in OpenSSH MD5 hash format
```
and
```
Fingerprint of the public key data in OpenSSH SHA256 hash format
```

detro: Does it make sense?

internal/openssh/lib.go
```diff
@@ -0,0 +1,214 @@
...
+
+	"golang.org/x/crypto/ssh"
+)
+
+// This module cherry-picks from https://go-review.googlesource.com/c/crypto/+/218620,
+// that provides a solution for https://github.com/golang/go/issues/37132, that highlights
+// the need for methods to marshal private keys to the OpenSSH format in `x/crypto/ssh`.
+//
+// This provides utilities to serialize a Private Keys in OpenSSH format: once
+// this goes mainstream, we can remove this module.
```
bflad, line 24: Nit: Can we please create the followup GitHub issue for tracking and reference it here please?

internal/provider/resource_private_key.go
```diff
@@ -125,26 +166,50 @@ func CreatePrivateKey(d *schema.ResourceData, meta interface{}) error {
...
+	d.Set("private_key_openssh", "")
+	if doMarshallOpenSSHKeyPemBlock {
+		openSSHKeyPemBlock, err := openssh.MarshalPrivateKey(key, "")
+		if err != nil {
+			return err
```
bflad, line 199: Nit: Let's wrap the error here so its clearer for practitioners about what happened

```suggestion
			return fmt.Errorf("unable to marshal private key into OpenSSH format: %w", err)
```

internal/provider/resource_private_key_test.go
```diff
@@ -15,36 +15,45 @@ func TestPrivateKeyRSA(t *testing.T) {
...
-                        value = "${tls_private_key.test.public_key_openssh}"
-                    }
-                    output "public_key_fingerprint_md5" {
-                        value = "${tls_private_key.test.public_key_fingerprint_md5}"
-                    }
-                `,
+					resource "tls_private_key" "test" {
+						algorithm = "RSA"
+					}
+					output "private_key_pem" {
```
bflad, line 21: Ah, this resource is using legacy style testing -- it might be worth updating it to use the helper/resource TestCheckFunc, such as [`TestMatchResourceAttr()`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource#TestMatchResourceAttr) directly against the resource attributes rather than creating outputs and referencing the state directly.

For example:

```go
Check: resource.ComposeAggregateTestCheckFunc(
  resource.TestMatchResourceAttr("tls_private_key.test", "private_key_pem", regexp.MustCompile(`^-----BEGIN OPENSSH PRIVATE KEY-----`)),
  // ... other TestCheckFunc ...
),
```

Refer to the [Terraform website](https://www.terraform.io/plugin/sdkv2/testing/acceptance-tests/teststep#builtin-check-functions) documentation for more details.

website/docs/r/private_key.html.md
```diff
@@ -24,43 +23,71 @@ state and does not create any external managed resources.
 ## Example Usage
 
 ```hcl
+// ECDSA key with P384 elliptic curve
```
bflad, line 26: Nit: Within Terraform HCL, `#` is more common for comments. Refer also to [Terraform configuration language](https://www.terraform.io/language/syntax/configuration#comments) documentation:

> The # single-line comment style is the default comment style and should be used in most cases. Automatic configuration formatting tools may automatically transform // comments into # comments, since the double-slash style is not idiomatic.

```suggestion
# ECDSA key with P384 elliptic curve
```

website/docs/r/private_key.html.md
```diff
@@ -24,43 +23,71 @@ state and does not create any external managed resources.
...
 resource "tls_private_key" "example" {
   algorithm   = "ECDSA"
   ecdsa_curve = "P384"
 }
+
+// RSA key of size 4096 bits
```
bflad, line 32:
```suggestion
```

website/docs/r/private_key.html.md
```diff
@@ -24,43 +23,71 @@ state and does not create any external managed resources.
...
+resource "tls_private_key" "example" {
```
bflad, line 33: Having multiple resources with the same name will generate an error if this whole configuration happens to be copy-pasted -- rather than using comments, the resource name labels could be adjusted to denote the different algorithms

```suggestion
resource "tls_private_key" "rsa-4096-example" {
```

website/docs/r/private_key.html.md
```diff
@@ -24,43 +23,71 @@ state and does not create any external managed resources.
...
+  algorithm = "RSA"
+  rsa_bits  = 4096
+}
+
+// ED25519 key
+resource "tls_private_key" "example" {
```
bflad, line 39:
```suggestion
resource "tls_private_key" "ed25519-example" {
```

website/docs/r/private_key.html.md
```diff
@@ -24,43 +23,71 @@ state and does not create any external managed resources.
...
 ```
 
 ## Argument Reference
 
 The following arguments are supported:
 
-* `algorithm` - (Required) The name of the algorithm to use for
-the key. Currently-supported values are "RSA" and "ECDSA".
+* `algorithm` - (Required) The name of the algorithm to use when generating
+the key. Currently-supported values are `RSA`, `ECDSA` and `ED25519`.
```
bflad, line 49: The algorithm names can be enforced at validation-time by using the [`helper/validation.StringInSlice` function](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation#StringInSlice) on the schema attribute, e.g.

```go
			"algorithm": {
				Type:        schema.TypeString,
				Required:    true,
				Description: "Name of the algorithm to use to generate the private key",
				ForceNew:    true,
				ValidateFunc: validation.StringInSlice([]string{ECDSA, ED25519, RSA}),
			},
```

website/docs/r/private_key.html.md
```diff
@@ -24,43 +23,71 @@ state and does not create any external managed resources.
...
+
+* `rsa_bits` - (Optional) When `algorithm` is `RSA`, the size of the generated
+RSA key in bits. Defaults to `2048`.
 
-* `rsa_bits` - (Optional) When `algorithm` is "RSA", the size of the generated
-RSA key in bits. Defaults to 2048.
+* `ecdsa_curve` - (Optional) When `algorithm` is `ECDSA`, the name of the elliptic
```
bflad, line 54: Similar validation can be done with this attribute

bflad (commented):
Looking good so far 😄 Just a few things to consider while we are in the neighborhood.

internal/provider/resource_private_key.go
kmoe: Yes, good wording

internal/openssh/lib.go
```diff
@@ -0,0 +1,214 @@
...
+	KeyAlgoED25519 = "ssh-ed25519"
+)
+
+const magic = "openssh-key-v1\x00"
+
+// MarshalPrivateKey returns a *pem.Block with the private key serialized in the OpenSSH format.
+//
+// See: https://coolaj86.com/articles/the-openssh-private-key-format/
+//
+// NOTE: `x/crypto/ssh` doesn't handle elliptic curve P-224, and so this applies to this function as well.
```
cipherboy, line 47: nit: I think its actually more fundamental that OpenSSH and the SSH protocol don't handle P-224 either. Outside of X.509, it isn't that widely used curve as P-256 and P-521 are fast enough for 99% of use cases and likely provide additional security.

internal/provider/resource_private_key.go
```diff
@@ -71,49 +90,71 @@ func resourcePrivateKey() *schema.Resource {
 			"ecdsa_curve": {
 				Type:        schema.TypeString,
 				Optional:    true,
-				Description: "ECDSA curve to use when generating a key",
+				Description: "Curve to use when generating an ECDSA key",
 				ForceNew:    true,
 				Default:     "P224",
```
cipherboy, line 95: Probably can't change it now, but P-256 is probably a saner default given the restrictions around SSH protocol not supporting P-224. :-)

website/docs/r/private_key.html.md
```diff
@@ -24,43 +23,71 @@ state and does not create any external managed resources.
...
+curve to use. Currently-supported values are `P224` (default), `P256`, `P384` or `P521`.
 
-* `ecdsa_curve` - (Optional) When `algorithm` is "ECDSA", the name of the elliptic
-curve to use. May be any one of "P224", "P256", "P384" or "P521", with "P224" as the
-default.
+~> **Limitations when using `ECDSA` with `P224` curve**
+When using `ECDSA` with `P224`, `.public_key_openssh`, `.private_key_openssh`, `.public_key_fingerprint_md5` and `.public_key_fingerprint_sha256` will have a value of `""`.
```
cipherboy, line 58: Ah, you might want to mention to see attribute reference below (due to inverse ordering). I wasn't sure why they'd have an empty fingerprint given that the PKIX key would be generated, but this is a _SSH_ fingerprint, not a [SKID/x.509 fingerprint](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.2) (which is probably not worthwhile to generate).

website/docs/r/private_key.html.md
```diff
@@ -3,14 +3,13 @@ layout: "tls"
...
 
 # tls\_private\_key
 
-Generates a secure private key and encodes it as PEM. This resource is
-primarily intended for easily bootstrapping throwaway development
-environments.
+Generates a secure private key and encodes it as [PEM (RFC 1421)](https://datatracker.ietf.org/doc/html/rfc1421).
+This resource is primarily intended for easily bootstrapping throwaway development environments.
 
 ~> **Important Security Notice** The private key generated by this resource will
```
cipherboy, line 14: Another note I'd probably add is that you shouldn't reuse the same key (output of this resource) between OpenSSH and TLS -- generate unique keys for each. Off the top of my head I'm not sure if its possible, but regardless, you don't want to open up the possibility of a cross-protocol attack due to key reuse (where e.g., you sign a TLS handshake hash or X.509 leaf cert with your OpenSSH private key that can then be massaged by an attacker into the correct form for impersonation in TLS/X.509 validation).

internal/openssh/lib.go
```diff
@@ -0,0 +1,214 @@
...
+// This module cherry-picks from https://go-review.googlesource.com/c/crypto/+/218620,
+// that provides a solution for https://github.com/golang/go/issues/37132, that highlights
+// the need for methods to marshal private keys to the OpenSSH format in `x/crypto/ssh`.
+//
+// This provides utilities to serialize a Private Keys in OpenSSH format: once
+// this goes mainstream, we can remove this module.
+
+// Constants represents the algorithm names for key types supported by this package.
+const (
+	KeyAlgoRSA = "ssh-rsa"
```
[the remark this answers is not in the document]
cipherboy: I guess it looks like you're only using them below in the imported code, so I'd probably suggest changing the references to `ssh.KeyAlgoRSA` &c below rather than re-defining them here.

cipherboy (commented):
Nothing major from my end, just a couple of nits.

internal/provider/resource_private_key.go
detro: yeah, discussed this yesterday with @kmoe too - it's very unfortunate, but changing it now will definitely break code that exists out there :(

internal/openssh/lib.go
detro: Good idea

internal/provider/resource_private_key_test.go
detro: Will try to refactor.

website/docs/r/private_key.html.md
detro: Silly me, of course! Will fix.

website/docs/r/private_key.html.md
detro: Oh, this is super nice!
Yes, absolutely!

internal/openssh/lib.go
detro: Will rephrase.

website/docs/r/private_key.html.md
detro: While this is very sound advice, doesn't it cross the line of "we give you the resource, then it's up to you what you do with it"?

I wouldn't want to turn the documentation of those resources into a "how to do cryptography with this stuff, correctly": it wouldn't fit and, additionally, it needs a proper expert to write it - I'm not one :D

internal/openssh/lib.go
detro: Created [this](https://github.com/hashicorp/terraform-provider-tls/issues/154): will add in the comment.

internal/provider/util.go
detro: I looked into switching to use this pattern, but the interfaced imposed by the SDK schema forces everything to end up a string anyway.

This is even more constrained by the fact that I'm adding the use of attribute validation in the schema:

```
ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(SupportedAlgorithmsString(), false)),
```

The alternative would be me writing validators dedicated to the `Algorithm` type, and it starts to feel like a lot of boilerplate. My goal was just to "nudge" future work away from having a bunch of copies of those strings around, and close the door on typos etc.

detro: PS Generics could helps with this kind of stuff a lot. Those validator functions we provide would be even more useful.

internal/provider/resource_private_key_test.go
detro: I didn't want to leave this without giving it a go:

```
{
Config: `
	resource "tls_private_key" "test" {
		algorithm = "RSA"
		rsa_bits = 4096
	}
	output "key_pem" {
		value = "${tls_private_key.test.private_key_pem}"
		sensitive = true
	}
`,
Check: r.ComposeAggregateTestCheckFunc(
	r.TestMatchResourceAttr("tls_private_key.test", "private_key_pem", regexp.MustCompile(`^-----BEGIN RSA PRIVATE KEY-----`)),
	func(s *terraform.State) error {
		pem := s.RootModule().Resources["tls_private_key.test"].Primary.Attributes["private_key_pem"]
		if len(pem) < 1700 {
			return fmt.Errorf("key PEM looks too short for a 4096-bit key (got %v characters)", len(pem))
		}
		return nil
	},
),
},
```

It definitely improves the length of tests, but I did find an example of function we might want to add to the `resource` package: something like a `TestResourceAttrLength()`. We could offer one that checks "string is between a range min/max", and another one "string length is".

detro: I just created this PR https://github.com/hashicorp/terraform-plugin-sdk/pull/893 that would help avoiding having to write that `func` in the example above to check the PEM length.

bookshelfdave (approved):

cipherboy (approved):
Looks like all my changes are addressed.

PR #140, opened 2021-12-23
fix unit test for resource_private_key_test

dttung2905:
Hi team,

I would like to put forward this MR to fix the error when running `make test`. 

```
❯ make test
==> Checking that code complies with gofmt requirements...
go test -i $(go list ./...) || exit 1
echo $(go list ./...) | \
        xargs -t -n4 go test  -timeout=120s -parallel=4
go test '-timeout=120s' '-parallel=4' github.com/terraform-providers/terraform-provider-tls github.com/terraform-providers/terraform-provider-tls/internal/provider 
?       github.com/terraform-providers/terraform-provider-tls   [no test files]
--- FAIL: TestPrivateKeyRSA (0.25s)
    resource_private_key_test.go:13: Step 1/2 error: Error running pre-apply refresh: exit status 1
        
        Error: Output refers to sensitive values
        
          on terraform_plugin_test.tf line 5:
           5:                     output "private_key_pem" {
        
        To reduce the risk of accidentally exporting sensitive data that was intended
        to be only internal, Terraform requires that any root module output
        containing sensitive data be explicitly marked as sensitive, to confirm your
        intent.
        
        If you do intend to export this data, annotate the output value as sensitive
        by adding the following argument:
            sensitive = true
--- FAIL: TestPrivateKeyECDSA (0.26s)
    resource_private_key_test.go:108: Step 1/2 error: Error running pre-apply refresh: exit status 1
        
        Error: Output refers to sensitive values
        
          on terraform_plugin_test.tf line 5:
           5:                     output "private_key_pem" {
        
        To reduce the risk of accidentally exporting sensitive data that was intended
        to be only internal, Terraform requires that any root module output
        containing sensitive data be explicitly marked as sensitive, to confirm your
        intent.
        
        If you do intend to export this data, annotate the output value as sensitive
        by adding the following argument:
            sensitive = true
FAIL
FAIL    github.com/terraform-providers/terraform-provider-tls/internal/provider 22.610s
FAIL
make: *** [GNUmakefile:12: test] Error 123

```

PR #153, opened 2022-02-21
Run acceptance tests against all the current minor versions of Terraform >= 0.12

detro:

Closes: #127


.github/workflows/test.yml
```diff
@@ -51,9 +51,12 @@ jobs:
       fail-fast: false
       matrix:
         terraform:
-          - '0.12.30'
-          - '0.13.6'
-          - '0.14.5'
+          - '0.12.31'
```
bflad, line 54: Nit: If we switch to using the `hashicorp/setup-terraform@v1` action instead of `TF_ACC_TERRAFORM_VERSION` environment variable we can install the latest patch versions. Refer to [terraform-provider-corner](https://github.com/hashicorp/terraform-provider-corner/blob/507647c80582ff6f4e4d3d81536db5906419e037/.github/workflows/test.yml#L48-L66) for an example. 😄

bflad (approved):
Looks good to me, just one thing to possibly consider. 🚀

.github/workflows/test.yml
detro: yeah, actually I have seen your PR yesterday and I think we should do the switch.

Having the flexibility to test against the latest patch and not having to manually maintain this list is the best way.

Will update this PR.

PS One last review before holidays? ;)

bryantbiggs:
FYI - I think this PR will also close #127

detro:
Thank you @bryantbiggs - I looked up but I couldn't find a PR or issue that described this issue.

I'll apply the changes suggested by @bflad and will link and close that one.

detro:
Closes #140

PR #155, opened 2022-02-24
QoL: Enabling Golang static code analysis with `golangci-lint`

detro:

## Description

I'm setting up golangci-lint with the all the defaults enabled, plus some extras. Follows a list of errors that were fixed, separated by "what caught them".

### Default linters (defined [here](https://golangci-lint.run/usage/linters/#enabled-by-default-linters))

```
internal/provider/data_source_public_key.go:82:7: Error return value of `d.Set` is not checked (errcheck)
	d.Set("algorithm", keyAlgo)
	     ^
internal/provider/resource_cert_request.go:137:7: Error return value of `d.Set` is not checked (errcheck)
	d.Set("cert_request_pem", certReqPem)
	     ^
internal/provider/resource_certificate.go:199:7: Error return value of `d.Set` is not checked (errcheck)
	d.Set("cert_pem", certPem)
	     ^
internal/openssh/lib_test.go:29:17: ineffectual assignment to err (ineffassign)
	rawPrivateKey, err := ssh.ParseRawPrivateKey(pemOpenSSHPrvKeyBytes)
	               ^
internal/openssh/lib_test.go:61:17: ineffectual assignment to err (ineffassign)
	rawPrivateKey, err := ssh.ParseRawPrivateKey(pemOpenSSHPrvKeyBytes)
	               ^
internal/openssh/lib_test.go:88:17: ineffectual assignment to err (ineffassign)
	rawPrivateKey, err := ssh.ParseRawPrivateKey(pemOpenSSHPrvKeyBytes)
	               ^
internal/provider/data_source_public_key.go:56:2: SA4006: this value of `bytes` is never used (staticcheck)
	bytes := []byte("")
...
```

### `deadcode, exportloopref, godot, gofmt, gosimple, govet, ineffassign, makezero, misspell, nilerr, staticcheck, structcheck, unconvert, unparam, unused, varcheck`

```
internal/provider/resource_private_key.go:21:22: Comment should end in a period (godot)
// selected algorithm
                     ^
internal/provider/resource_private_key.go:25:39: Comment should end in a period (godot)
// according to the selected algorithm
                                      ^
internal/provider/types.go:3:56: Comment should end in a period (godot)
// Algorithm represents a type of private key algorithm
                                                       ^
internal/provider/resource_certificate.go:135: File is not `gofmt`-ed with `-s` (gofmt)
		"set_subject_key_id": &schema.Schema{
internal/provider/provider.go:35:72: nameFromResourceData - result 1 (error) is always nil (unparam)
func nameFromResourceData(nameMap map[string]interface{}) (*pkix.Name, error) {
                                                                       ^
```

### `predeclared, durationcheck, errcheck`

```
internal/provider/resource_certificate.go:148:45: Multiplication of durations: `time.Duration(d.Get("validity_period_hours").(int)) * time.Hour` (durationcheck)
	template.NotAfter = template.NotBefore.Add(time.Duration(d.Get("validity_period_hours").(int)) * time.Hour)
	                                           ^
internal/provider/resource_certificate.go:234:25: Multiplication of durations: `time.Duration(-d.Get("early_renewal_hours").(int)) * time.Hour` (durationcheck)
		earlyRenewalPeriod := time.Duration(-d.Get("early_renewal_hours").(int)) * time.Hour
```

## Links

* https://golangci-lint.run/usage/linters/
* https://github.com/hashicorp/terraform-plugin-sdk/blob/main/.golangci.yml

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,17 @@
+linters:
+  # Default linters enabled
+  # See: https://golangci-lint.run/usage/linters/#enabled-by-default-linters
+
+  # Additional linters enabled
+  enable:
+    - durationcheck
+    - exportloopref
+    - godot
+    - gofmt
+    - makezero
+    - misspell
+    - nilerr
+    - predeclared
+#    - tenv # TODO: Enable when upgrading Go 1.16 to 1.17
+    - unconvert
+    - unparam
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -32,7 +32,7 @@ func hashForState(value string) string {
 	return hex.EncodeToString(hash[:])
 }
 
-func nameFromResourceData(nameMap map[string]interface{}) (*pkix.Name, error) {
+func nameFromResourceData(nameMap map[string]interface{}) *pkix.Name {
 	result := &pkix.Name{}
 
 	if value := nameMap["common_name"]; value != "" {
@@ -66,7 +66,7 @@ func nameFromResourceData(nameMap map[string]interface{}) (*pkix.Name, error) {
 		result.SerialNumber = value.(string)
 	}
 
-	return result, nil
+	return result
 }
 
 var nameSchema *schema.Resource = &schema.Resource{
--- a/internal/provider/util.go
+++ b/internal/provider/util.go
@@ -88,20 +88,34 @@ func readPublicKey(d *schema.ResourceData, prvKey interface{}) error {
 	}
 
 	d.SetId(hashForState(string(pubKeyBytes)))
-	d.Set("public_key_pem", string(pem.EncodeToMemory(pubKeyPemBlock)))
+
+	if err := d.Set("public_key_pem", string(pem.EncodeToMemory(pubKeyPemBlock))); err != nil {
+		return fmt.Errorf("error setting value on key 'public_key_pem': %s", err)
+	}
 
 	// NOTE: ECDSA keys with elliptic curve P-224 are not supported by `x/crypto/ssh`,
 	// so this will return an error: in that case, we set the below fields to emptry strings
 	sshPubKey, err := ssh.NewPublicKey(publicKey(prvKey))
+	var pubKeySSH, pubKeySSHFingerprintMD5, pubKeySSHFingerprintSHA256 string
 	if err == nil {
 		sshPubKeyBytes := ssh.MarshalAuthorizedKey(sshPubKey)
-		d.Set("public_key_openssh", string(sshPubKeyBytes))
-		d.Set("public_key_fingerprint_md5", ssh.FingerprintLegacyMD5(sshPubKey))
-		d.Set("public_key_fingerprint_sha256", ssh.FingerprintSHA256(sshPubKey))
-	} else {
-		d.Set("public_key_openssh", "")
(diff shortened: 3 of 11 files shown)


internal/provider/util.go
```diff
@@ -88,20 +88,34 @@ func readPublicKey(d *schema.ResourceData, prvKey interface{}) error {
...
-		d.Set("public_key_openssh", "")
-		d.Set("public_key_fingerprint_md5", "")
-		d.Set("public_key_fingerprint_sha256", "")
+
+		pubKeySSH = string(sshPubKeyBytes)
+		pubKeySSHFingerprintMD5 = ssh.FingerprintLegacyMD5(sshPubKey)
+		pubKeySSHFingerprintSHA256 = ssh.FingerprintSHA256(sshPubKey)
+	}
+
+	if err := d.Set("public_key_openssh", pubKeySSH); err != nil {
```
bookshelfdave, line 108: In the original `else` block, all values are set to empty strings. Is that behavior still needed? (my guess is no)

internal/provider/provider.go
```diff
@@ -66,7 +66,7 @@ func nameFromResourceData(nameMap map[string]interface{}) (*pkix.Name, error) {
 		result.SerialNumber = value.(string)
 	}
 
-	return result, nil
+	return result
```
bookshelfdave, line 69: When `if` is false above, does the initialized/empty `result` signify an error somehow?

bookshelfdave (approved):
lgtm 👍 , just a few misc questions

.golangci.yml
```diff
@@ -0,0 +1,17 @@
...
+  enable:
+    - durationcheck
+    - exportloopref
+    - godot
+    - gofmt
+    - makezero
+    - misspell
+    - nilerr
+    - predeclared
+#    - tenv # TODO: Enable when upgrading Go 1.16 to 1.17
```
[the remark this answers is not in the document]
detro: Added it [here](https://github.com/hashicorp/terraform-providers-devex-internal/issues/94) and [here](https://docs.google.com/spreadsheets/d/1eI7iGRxw1tdAg96Kn4Oomfgd9SBDWOxMx4g7Ah3LVCM/edit#gid=0)

internal/provider/provider.go
detro: I think the signature was written BEFORE the implementation was put down.

In a certificate, all those fields are optional, so it return an empty `result` and it should be ok.

internal/provider/util.go
detro: So, I wanted to avoid having to write the same `if err := d.Set...` 6 times, was kinda ugly. And the static analysis reported that we were not checking the return error from `.Set`.

So, I have initialized `publicKeySSH*` variables to empty strings: if they get a value (i.e. `ssh.NewPublicKey()` does not error), we set those strings to a proper value. Otherwise, they are already empty string.

This preserves the previous behaviour WHILE making sure we call `.Set` for each attribute only once.

bflad: This looks good -- its important to preserve the `""` setting behavior where it currently exists since not doing so is an odd backwards compatibility issue until a next major version (which we may want to consider breaking).

PR #156, opened 2022-03-01
Upgrading to Golang 1.17

detro:

As part of this, I have also enabled `tenv` [provided](https://golangci-lint.run/usage/linters/#tenv) by GolangCi Lint.


bflad (approved):
Looks good to me! 🚀

PR #169, opened 2022-03-17
QoL: Updating all the provider Golang code to include `Description` in preparation of documentation generation

detro:
This is extrapolated by #167, but makes the change-set more "manageable" for a good review process.

Additional to the addition of a lot of markdown-based `Description` strings to the `Schema`s, this slightly re-organises the common part of the `*_cert.go` resources, so their schemas are set by the same utility methods (see `common_cert.go`).

I have explicitly used `git mv resource_certificate.go common_cert.go` but somehow git decided it's still not a move and created a delete and a create 😢


internal/provider/common_cert.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+
+	hash := sha1.Sum(publicKeyBytes)
+	return hash[:], nil
+}
+
+// setCertificateSubjectSchema sets on the given reference to map of schema.Schema
+// all the keys required by a resource representing a certificate's subject.
+func setCertificateSubjectSchema(s *map[string]*schema.Schema) {
+	(*s)["dns_names"] = &schema.Schema{
```
bflad, line 94: Nit: Using map pointers is extraneous -- these can be simplified to 

```go
func setCertificateSubjectSchema(s map[string]*schema.Schema) {
	s["dns_names"] = &schema.Schema{
```

etc. 👍

internal/provider/common_cert.go
```diff
@@ -0,0 +1,463 @@
...
+		Description: "The subject for which a certificate is being requested. " +
+			"The acceptable arguments are all optional and their naming is based upon " +
+			"[Issuer Distinguished Names (RFC5280)](https://tools.ietf.org/html/rfc5280#section-4.1.2.4) section.",
+	}
+}
+
+// setCertificateCommonSchema sets on the given reference to map of schema.Schema
+// all the keys required by a resource representing a certificate.
+func setCertificateCommonSchema(s *map[string]*schema.Schema) {
+	(*s)["validity_period_hours"] = &schema.Schema{
```
bflad, line 220: Similarly

```go
func setCertificateCommonSchema(s map[string]*schema.Schema) {
	s["validity_period_hours"] = &schema.Schema{
```

bflad (approved):
Some minor considerations, otherwise looks good to me 🚀

internal/provider/common_cert.go
detro: Yeah, I forgot to integrate that feedback. Will do now.

internal/provider/common_cert.go
```diff
@@ -0,0 +1,463 @@
...
+		Description: "Is the generated certificate representing a Certificate Authority (CA) (default: `false`).",
+	}
+
+	(*s)["allowed_uses"] = &schema.Schema{
+		Type:     schema.TypeList,
+		Required: true,
+		ForceNew: true,
+		Elem: &schema.Schema{
+			Type:         schema.TypeString,
+			ValidateFunc: validation.StringInSlice(supportedKeyUsages(), false),
```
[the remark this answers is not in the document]
detro: Yes, I think it's worth starting populating the CHANGELOG for 3.2.0.

PR #170, opened 2022-03-18
QoL: wiring up `tfplugindocs` to the code generation

detro:

Follow up from: #169


GNUmakefile
```diff
@@ -1,45 +1,26 @@
-TEST?=$$(go list ./...)
-GOFMT_FILES?=$$(find . -name '*.go')
-PKG_NAME=tls
+GOLANGCILINT_INSTALLED := $(shell command -v golangci-lint 2> /dev/null)
```
bflad, line 1: Open question: Is this Windows compatible? If not, let's just let the underlying shell interpreter bubble its own "command not found" like error.

bflad (approved):
One quick question, otherwise looks good to me 🚀

GNUmakefile
detro: That's a good point.

Until now our makefile for sure couldn't work on Windows (the use of `find`, `xargs` etc. make it impossible).

With this revamp, there is actually a chance that it might work on Windows (bar the  `make` dependency).

I'll get rid of this: a "command not found" would deliver pretty much the same thing.
I'll just leave a link to `golangci-lint` to help future contributors.

PR #171, opened 2022-03-21
QoL: Regenerated doc using `tfplugindocs`

detro:
Follow up from: #170 and #169 

This (last) PR make use of what we did so far with `tfplugindocs` in this project and finally replaces the manually maintained documentation with the auto-generated one.

I managed to preserve the history of the documentation files we are now generating, so it should be much easier to spot differences and issues.


bflad (approved):
Looks good to me

PR #173, opened 2022-03-21
Adding support for ED25519 to all the `*_cert` resources

detro:
Context: #150

By moving all the utilities related to `*_cert` in `common_cert.go`, and doing the same for `*.key`, it is now easy to have the same support for Key Algorithm across the board.

With this, all the certs support ED25519 private keys, without any code specific for this in their implementation.

Additionally, the fields `*_key_algorithm` used in the `*_cert` resources are now all marked as `Optional` and `Deprecated`: the plan is to release 3.2.0 first, then change those fields to `Computed` (i.e. exposing the inferred algorithm), and then release a 4.0.0. This (should be) in accordance with our guidelines: https://www.terraform.io/plugin/sdkv2/best-practices/deprecations.

#### Today 

```
"key_algorithm": &schema.Schema{
  Type:     schema.TypeString,
  Required: true,
  ForceNew: true,
```

#### 3.2.0 (i.e. this PR)

```
"key_algorithm": &schema.Schema{
  Type:       schema.TypeString,
  Optional:   true,
  ForceNew:   true,
  Deprecated: "DEPRECATION MESSAGE"
```

This will:

1. inform user with a warning that the field should not be used anymore
2. if the user complies, the state gets updated with the removal of the string from it

At this point, we release 3.2.0

#### 4.0.0 (future PR)

```
"key_algorithm": &schema.Schema{
  Type:       schema.TypeString,
  Computed:   true,
```

This will:

1. allow the provider to `.Set("key_algorithm")`
2. the state is updated with the computed value

### DIFF
--- a/internal/provider/common_cert.go
+++ b/internal/provider/common_cert.go
@@ -461,3 +461,52 @@ func distinguishedNamesFromSubjectAttributes(nameMap map[string]interface{}) *pk
 
 	return result
 }
+
+func parseCertificate(d *schema.ResourceData, pemKey string) (*x509.Certificate, error) {
+	block, err := decodePEM(d, pemKey, "")
+	if err != nil {
+		return nil, err
+	}
+
+	certs, err := x509.ParseCertificates(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse %s: %s", pemKey, err)
+	}
+	if len(certs) < 1 {
+		return nil, fmt.Errorf("no certificates found in %s", pemKey)
+	}
+	if len(certs) > 1 {
+		return nil, fmt.Errorf("multiple certificates found in %s", pemKey)
+	}
+
+	return certs[0], nil
+}
+
+func parseCertificateRequest(d *schema.ResourceData, pemKey string) (*x509.CertificateRequest, error) {
+	block, err := decodePEM(d, pemKey, CertificateRequest.String())
+	if err != nil {
+		return nil, err
+	}
+
+	certReq, err := x509.ParseCertificateRequest(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse %s: %s", pemKey, err)
+	}
+
+	return certReq, nil
+}
+
+func certificateToMap(cert *x509.Certificate) map[string]interface{} {
+	return map[string]interface{}{
+		"signature_algorithm":  cert.SignatureAlgorithm.String(),
+		"public_key_algorithm": cert.PublicKeyAlgorithm.String(),
+		"serial_number":        cert.SerialNumber.String(),
+		"is_ca":                cert.IsCA,
+		"version":              cert.Version,
+		"issuer":               cert.Issuer.String(),
+		"subject":              cert.Subject.String(),
+		"not_before":           cert.NotBefore.Format(time.RFC3339),
+		"not_after":            cert.NotAfter.Format(time.RFC3339),
+		"sha1_fingerprint":     fmt.Sprintf("%x", sha1.Sum(cert.Raw)),
+	}
+}
--- a/internal/provider/common_cert.go
+++ b/internal/provider/common_cert.go
@@ -349,7 +349,7 @@ func createCertificate(d *schema.ResourceData, template, parent *x509.Certificat
 	if err != nil {
 		return fmt.Errorf("error creating certificate: %s", err)
 	}
-	certPem := string(pem.EncodeToMemory(&pem.Block{Type: Certificate.String(), Bytes: certBytes}))
+	certPem := string(pem.EncodeToMemory(&pem.Block{Type: PreambleCertificate.String(), Bytes: certBytes}))
 
 	validFromBytes, err := template.NotBefore.MarshalText()
 	if err != nil {
@@ -485,7 +485,7 @@ func parseCertificate(d *schema.ResourceData, pemKey string) (*x509.Certificate,
 }
 
 func parseCertificateRequest(d *schema.ResourceData, pemKey string) (*x509.CertificateRequest, error) {
-	block, err := decodePEM(d, pemKey, CertificateRequest.String())
+	block, err := decodePEM(d, pemKey, PreambleCertificateRequest.String())
 	if err != nil {
 		return nil, err
 	}
--- a/internal/provider/common_cert.go
+++ b/internal/provider/common_cert.go
@@ -122,10 +122,14 @@ func setCertificateSubjectSchema(s map[string]*schema.Schema) {
 	}
 
 	s["key_algorithm"] = &schema.Schema{
-		Type:        schema.TypeString,
-		Required:    true,
-		ForceNew:    true,
-		Description: "Name of the algorithm used when generating the private key provided in `private_key_pem`.",
+		Type:     schema.TypeString,
+		Optional: true,
+		ForceNew: true,
+		Deprecated: "This is now ignored, as the key algorithm is inferred from the `private_key_pem`. " +
+			"It it will be made read-only in the next major release.",
+		Description: "Name of the algorithm used when generating the private key provided in `private_key_pem`. " +
+			"**NOTE**: this is deprecated and ignored, as the key algorithm is now inferred from the key. " +
+			"It it will be made read-only in the next major release.",
 	}
 
 	s["private_key_pem"] = &schema.Schema{
--- a/internal/provider/common_cert.go
+++ b/internal/provider/common_cert.go
@@ -124,6 +124,7 @@ func setCertificateSubjectSchema(s map[string]*schema.Schema) {
 	s["key_algorithm"] = &schema.Schema{
 		Type:       schema.TypeString,
 		Optional:   true,
+		Computed:   true,
 		ForceNew:   true,
 		Deprecated: "This is now ignored, as the key algorithm is inferred from the `private_key_pem`.",
 		Description: "Name of the algorithm used when generating the private key provided in `private_key_pem`. " +
--- a/internal/provider/common_cert.go
+++ b/internal/provider/common_cert.go
@@ -122,14 +122,12 @@ func setCertificateSubjectSchema(s map[string]*schema.Schema) {
 	}
 
 	s["key_algorithm"] = &schema.Schema{
-		Type:     schema.TypeString,
-		Optional: true,
-		ForceNew: true,
-		Deprecated: "This is now ignored, as the key algorithm is inferred from the `private_key_pem`. " +
-			"It it will be made read-only in the next major release.",
+		Type:       schema.TypeString,
+		Optional:   true,
+		ForceNew:   true,
+		Deprecated: "This is now ignored, as the key algorithm is inferred from the `private_key_pem`.",
 		Description: "Name of the algorithm used when generating the private key provided in `private_key_pem`. " +
-			"**NOTE**: this is deprecated and ignored, as the key algorithm is now inferred from the key. " +
-			"It it will be made read-only in the next major release.",
+			"**NOTE**: this is deprecated and ignored, as the key algorithm is now inferred from the key. ",
 	}
 
 	s["private_key_pem"] = &schema.Schema{
--- a/internal/provider/common_key.go
+++ b/internal/provider/common_key.go
@@ -1,0 +1,136 @@
+package provider
+
+import (
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"golang.org/x/crypto/ssh"
+)
+
+// parsePrivateKeyPEM takes a slide of bytes containing a private key
+// encoded in [PEM (RFC 1421)](https://datatracker.ietf.org/doc/html/rfc1421) format,
+// and returns a crypto.PrivateKey implementation, together with the Algorithm used by the key.
+func parsePrivateKeyPEM(keyPEMBytes []byte) (crypto.PrivateKey, Algorithm, error) {
+	pemBlock, rest := pem.Decode(keyPEMBytes)
+	if pemBlock == nil {
+		return nil, "", fmt.Errorf("failed to decode PEM block: decoded bytes %d, undecoded %d", len(keyPEMBytes)-len(rest), len(rest))
+	}
+
+	// Map PEM Preamble of the Private Key to the corresponding Algorithm
+	algorithm, err := PEMPreamblePrivateKey(pemBlock.Type).Algorithm()
+	if err != nil {
+		return nil, "", err
+	}
+
+	// Identify parser for the given key, using the Algorithm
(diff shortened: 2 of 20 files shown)


internal/provider/common_key.go
```diff
@@ -0,0 +1,190 @@
...
+	},
+	ED25519: func(der []byte) (crypto.PrivateKey, error) {
+		return x509.ParsePKCS8PrivateKey(der)
+	},
+}
+
+// parsePrivateKeyPEM takes a slide of bytes containing a private key
+// encoded in [PEM (RFC 1421)](https://datatracker.ietf.org/doc/html/rfc1421) format,
+// and returns a crypto.PrivateKey implementation, together with the Algorithm used by the key.
+func parsePrivateKeyPEM(keyPEMBytes []byte) (crypto.PrivateKey, Algorithm, error) {
```
detro, line 73: This is _net new_ in this file, and now the 1 function used to parse ALL PEM private keys across the provider.

internal/provider/common_key.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil, "", fmt.Errorf("failed to parse private key of algorithm %s: %w", algorithm, err)
+	}
+
+	return prvKey, algorithm, nil
+}
+
+// parsePrivateKeyOpenSSHPEM takes a slide of bytes containing a private key
+// encoded in [OpenSSH PEM (RFC 4716)](https://datatracker.ietf.org/doc/html/rfc4716) format,
+// and returns a crypto.PrivateKey implementation, together with the Algorithm used by the key.
+func parsePrivateKeyOpenSSHPEM(keyOpenSSHPEMBytes []byte) (crypto.PrivateKey, Algorithm, error) {
```
detro, line 103: This is _net new_ in this file, and now the 1 function used to parse ALL OpenSSH PEM private keys across the provider.

detro:
ping @bflad :)

bflad (changes requested):
I think this is generally moving in an okay direction, but I have left some testing and security considerations. It may be good to reach out to additional security folks in this situation to more authoritatively determine the safety of this change (where still considering an "expected" algorithm may be a decent compromise).

internal/provider/resource_cert_request_test.go
```diff
@@ -45,7 +45,6 @@ func TestCertRequest(t *testing.T) {
                             "spiffe://example-trust-domain/workload2",
                         ]
 
-                        key_algorithm = "RSA"
```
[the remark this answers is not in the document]
detro: Will have 2 tests for the before and after 👍

internal/provider/common_cert.go
```diff
@@ -122,10 +122,14 @@ func setCertificateSubjectSchema(s map[string]*schema.Schema) {
...
 	s["key_algorithm"] = &schema.Schema{
-		Type:        schema.TypeString,
-		Required:    true,
-		ForceNew:    true,
-		Description: "Name of the algorithm used when generating the private key provided in `private_key_pem`.",
+		Type:     schema.TypeString,
+		Optional: true,
+		ForceNew: true,
+		Deprecated: "This is now ignored, as the key algorithm is inferred from the `private_key_pem`. " +
+			"It it will be made read-only in the next major release.",
```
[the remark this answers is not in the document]
detro: In my testing I concluded that a 2 step approach (first make it `Optional`, release, then make it `Computed`) was necessary to avoid causing unnecessary re-creation of the resources.

I can remove the reference to "future behaviour" as what matters is that ppl stop using it.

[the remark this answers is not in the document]
bflad: I should've clarified better, sorry. I meant `Optional` and `Computed`, with calling `d.Set()`. `Optional` + `Computed` + `ForceNew` should not trigger recreation when removing the attribute from the configuration and if it does, that would be unexpected behavior we should investigate further.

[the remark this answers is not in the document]
detro: So, here is the issue with adding `Computer` to `Optional` now: it will be a breaking change.

I think we are expected to cross that bridge only once we release a major version, after a minor version with the deprecation - no?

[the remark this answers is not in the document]
detro: To be clear, I tried yesterday all the combinations, and just changing from `Required` to `Optional` for now, seemed like the only path forward to allow for an opt-in migration.

All other options would end up with a breaking change.

internal/provider/common_key.go
```diff
@@ -0,0 +1,190 @@
...
+// encoded in [PEM (RFC 1421)](https://datatracker.ietf.org/doc/html/rfc1421) format,
+// and returns a crypto.PrivateKey implementation, together with the Algorithm used by the key.
+func parsePrivateKeyPEM(keyPEMBytes []byte) (crypto.PrivateKey, Algorithm, error) {
+	pemBlock, rest := pem.Decode(keyPEMBytes)
+	if pemBlock == nil {
+		return nil, "", fmt.Errorf("failed to decode PEM block: decoded bytes %d, undecoded %d", len(keyPEMBytes)-len(rest), len(rest))
+	}
+
+	// Map PEM Preamble of the Private Key to the corresponding Algorithm
+	algorithm, err := PEMPreamblePrivateKey(pemBlock.Type).Algorithm()
```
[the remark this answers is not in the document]
detro: OK so, I have dug into the code of `crypto/ssh` and `crypto.x509` and the Preamble (i.e. the `pem.Block.Type` field) are indeed what drives the parsing decision in all the "switches".

Given the ambiguity in the RFC that I highlighted in our private chat, I feel like the best course of action is to imitate what Golang libraries already do.

I will of course integrate any feedback I might get back from the security team, so far nothing.

Additional to all of this, by poking me on this it made me realise that there was a corner case currently unhandled: the Preamble `BEGING PRIVATE KEY` can be used for more than just ED25519, but it can instead be used for RSA and ECDSA too.

Here is the signature of the method designated to parse it:

```golang
// ParsePKCS8PrivateKey parses an unencrypted private key in PKCS #8, ASN.1 DER form.
//
// It returns a *rsa.PrivateKey, a *ecdsa.PrivateKey, or a ed25519.PrivateKey.
// More types might be supported in the future.
//
// This kind of key is commonly encoded in PEM blocks of type "PRIVATE KEY".
func ParsePKCS8PrivateKey(der []byte) (key interface{}, err error) {
  // ...
}
```

So, I have restructured things to handle the key identification AFTER the key parsing.

internal/provider/common_cert.go
```diff
@@ -122,10 +122,12 @@ func setCertificateSubjectSchema(s map[string]*schema.Schema) {
 	}
 
 	s["key_algorithm"] = &schema.Schema{
-		Type:        schema.TypeString,
-		Required:    true,
-		ForceNew:    true,
-		Description: "Name of the algorithm used when generating the private key provided in `private_key_pem`.",
+		Type:       schema.TypeString,
+		Optional:   true,
```
bflad, line 126:
```suggestion
		Optional:   true,
		Computed:   true,
```

internal/provider/resource_locally_signed_cert.go
```diff
@@ -23,10 +23,12 @@ func resourceLocallySignedCert() *schema.Resource {
 	}
 
 	s["ca_key_algorithm"] = &schema.Schema{
-		Type:        schema.TypeString,
-		Required:    true,
-		ForceNew:    true,
-		Description: "Name of the algorithm used when generating the private key provided in `ca_private_key_pem`.",
+		Type:       schema.TypeString,
+		Optional:   true,
```
bflad, line 27:
```suggestion
		Optional:   true,
		Computed:   true,
```

internal/provider/resource_locally_signed_cert.go
```diff
@@ -70,7 +72,7 @@ func createLocallySignedCert(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return err
 	}
-	caKey, err := parsePrivateKey(d, "ca_private_key_pem", "ca_key_algorithm")
+	caKey, _, err := parsePrivateKeyPEM([]byte(d.Get("ca_private_key_pem").(string)))
```
bflad, line 75: Similar note about getting the algorithm here and calling `d.Set("ca_key_algorithm", caKeyAlgorithm)`

```suggestion
	caKey, caKeyAlgorithm, err := parsePrivateKeyPEM([]byte(d.Get("ca_private_key_pem").(string)))
```

internal/provider/resource_locally_signed_cert.go
```diff
@@ -70,7 +72,7 @@ func createLocallySignedCert(d *schema.ResourceData, meta interface{}) error {
...
 	if err != nil {
 		return err
 	}
```
bflad, line 78:
```suggestion
	}
	
	if err := d.Set("ca_key_algorithm", keyAlgorithm); err != nil {
		return fmt.Errorf("unable to set ca_key_algorithm: %w", err)
	}
```

internal/provider/resource_locally_signed_cert_test.go
```diff
@@ -297,7 +297,143 @@ func TestAccLocallySignedCertNotRecreatedForEarlyRenewalUpdateInFuture(t *testin
...
+			},
+		},
+	})
+}
+
+func locallySignedCertConfig(validity, earlyRenewal uint32, setKeyAlgorithm bool) string {
+	caKeyAlgorithmCfg := ""
+	if setKeyAlgorithm {
+		caKeyAlgorithmCfg = `ca_key_algorithm = "RSA"`
+	}
```
bflad, line 435: Rather than implementing this logic in Go it could also be done in the configuration itself with conditional logic, e.g.

```
ca_key_algorithm = %t ? "RSA" : null
```

Although admittedly, most providers prefer to create wholly separate configurations for clarity following the Go practice of preferring some duplication for clarity over being purely DRY. (It makes much more sense when you're working with dozens or hundreds of tests so you don't need to figure out any special testing configuration logic.)

bflad (changes requested):
Here's the full reproduction of how I'm coming to the conclusion about `Computed: true` and calling `d.Set()`. Please reach out if you need help getting this going in your local development environment.

Starting configuration (e.g. create a directory with this as `main.tf`):

```terraform
terraform {
  required_providers {
    tls = {
      source  = "hashicorp/tls"
      version = "3.1.0"
    }
  }

  required_version = "1.1.7"
}

resource "tls_private_key" "example" {
  algorithm   = "ECDSA"
  ecdsa_curve = "P384"
}

resource "tls_self_signed_cert" "example" {
  key_algorithm   = "ECDSA"
  private_key_pem = tls_private_key.example.private_key_pem

  subject {
    common_name  = "example.com"
    organization = "ACME Examples, Inc"
  }

  validity_period_hours = 12

  allowed_uses = [
    "key_encipherment",
    "digital_signature",
    "server_auth",
  ]
}

output "algorithm" {
  value = tls_self_signed_cert.example.key_algorithm
}
```

Install and use the production 3.1.0 provider to create a starting state that represents what a practitioner may have today:

```shell
terraform init
terraform apply -auto-approve
```

With that starting state in place, we can experiment with the updated provider locally outside the testing framework. Via this example `.terraformrc` file (e.g. I just created it in the configuration directory as `.terraformrc`; it should be updated for your `GOBIN`):

```text
provider_installation {
  dev_overrides {
    "hashicorp/tls" = "/Users/bflad/go/bin"
  }

  direct {}
}
```

In the provider code directory, run `go install .` from this PR branch to install the provider into the `GOBIN`.

Running `TF_CLI_CONFIG_FILE=./.terraformrc terraform apply` without configuration changes yields no relevant changes and a deprecation warning, which is great and what we expect.

If I then comment `key_algorithm   = "ECDSA"`in `tls_self_signed_cert` to try and fix the deprecation, running `TF_CLI_CONFIG_FILE=./.terraformrc terraform apply` yields this unexpected plan:

```text
      - key_algorithm         = "ECDSA" -> null # forces replacement
```

That is not what we are looking for here. I should be able to fix the deprecation in my configuration without other effects. If I update the provider schema to instead include `Computed: true` here:

```go
	s["key_algorithm"] = &schema.Schema{
		Type:       schema.TypeString,
		Optional:   true,
		Computed:   true,
		ForceNew:   true,
		Deprecated: "This is now ignored, as the key algorithm is inferred from the `private_key_pem`.",
		Description: "Name of the algorithm used when generating the private key provided in `private_key_pem`. " +
			"**NOTE**: this is deprecated and ignored, as the key algorithm is now inferred from the key. ",
	}
```

And update the `CreateSelfSignedCert` function to return the algorithm and call `d.Set()`:

```go
key, keyAlgorithm, err := parsePrivateKeyPEM([]byte(d.Get("private_key_pem").(string)))

if err != nil {
	return err
}

// Ensure key_algorithm is always set, even when not in the configuration
if err := d.Set("key_algorithm", keyAlgorithm); err != nil {
	return fmt.Errorf("unable to set key_algorithm: %w", err)
}
// ... rest of logic ...
```

With the schema and create function updates done, we can now call `go install .` to reinstall the local provider. Running `TF_CLI_CONFIG_FILE=./.terraformrc terraform apply` now yields no changes, which is what I'd expect.

Using this updated provider, we can also verify that it works as expected on new resource creation without configuring `key_algorithm` in the configuration:

```shell
rm -rf terraform.tfstate*
TF_CLI_CONFIG_FILE=./.terraformrc terraform apply -auto-approve
```

This also works as expected. We can use this same pattern for other similar resources.

---

Aside: Typically, we'd also update `readCertificate` to do the same so the algorithm is always refreshed (in case of import), however this schema implements `StateFunc` in a way we typically do not recommend so we cannot `d.Get()` the private_key_pem like normal from the `Read` function. The [[sensitive state page](https://www.terraform.io/plugin/sdkv2/best-practices/sensitive-state)](https://www.terraform.io/plugin/sdkv2/best-practices/sensitive-state) used to have a note about not hashing values, but I forget why it was removed. If we did want to support import in the future, the import function itself could extract and set the key_algorithm attribute appropriately though.

detro:
OK, one thing is clear: what the tests can't reproduce (easily) is the behaviour of such changes when there is a pre-existing state.

Not having `Computed: true` and the practitioner removing `key_algorithm`, triggers the change from value being set to `null`: 

```
- key_algorithm         = "ECDSA" -> null # forces replacement
```

The same does NOT happen when `Computed: true` is indeed set: the output in your example can STILL return the value `ECDSA`, even though in configuration is gone and there is no `d.Set()` in my code yet.

I have to presume that terraform, seeing that it's computed, it initializes the resource during the `plan` from the state, so we carry that forward.

(cont...)

internal/provider/resource_locally_signed_cert_test.go
detro: Given those tests are only temporarily there, is it ok if I punt on it? I intend to release 4.x soon after this.

detro:
OK @bflad , this should be ready for another review: I have set the fields to `Computed + Optional` + of course added calls to `d.Set("key_algorithm")` as expected.

internal/provider/resource_locally_signed_cert_test.go
bflad: Totally! Not a big deal, just wanted to let you know. 😄

bflad (approved):
Thank you for sticking through me with this and it looks good to go! 🚀

PR #113, opened 2021-04-29
dataSourceTlsCertificateRead now uses http client

DanielRis:
`dataSourceTlsCertificateRead` now use an http client instead of tls to allow proxy usage inside Corporate environments.

Fixes #96

Issue #174, opened 2022-03-25
Remove the deprecated `key_algorithm` argument from `*_cert` resources


detro:
### Terraform CLI and Provider Versions

0.12

### Use Cases or Problem Statement

Context: #173 

The Algorithm of the Private Key algorithm is inferred from the Key itself as of 3.2.0.
We have deprecated it in 3.2.0: this issue is about removing it, and replacing it with a `Computed` value that plays an "informative" role.

### Proposal

Fields `key_algorithm` are removed after deprecation in `3.2.0`.
Also testing setup to aid the deprecation should be removed.

### How much impact is this issue causing?

Low

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #179, opened 2022-03-31
Support for Proxy on the Provider level

### LINKED ISSUE #96: [feature request] Support HTTP proxy for tls_certificate data source
I am trying to retrieve a certificate via the `tls_certificate` data source as per the [example in the docs](https://registry.terraform.io/providers/hashicorp/tls/latest/docs/data-sources/tls_certificate#example-usage). This is in context of an OIDC provider for an EKS cluster running in a private VPC (_no public network connectivity_). 

Given an OIDC issuer URL such as `https://oidc.eks.eu-west-1.amazonaws.com/id/838DB18FA8EC26D87848FDEC1D50CFDC` the call evetually fails with:

```
Error: dial tcp 34.248.140.150:443: connect: connection timed out
```

This is because of the private VPC and the fact that the data source _does not support a HTTP proxy_ (there are unfortunately no EKS VPC [endpoints](https://docs.aws.amazon.com/vpc/latest/userguide/integrated-services-vpce-list.html) currently).

### Terraform Version
`0.14.4`

### Affected Data Source(s)
- `tls_certificate`

```hcl
data "tls_certificate" "eks_cluster_oidc_issuer" {
  // E.g. https://oidc.eks.eu-west-1.amazonaws.com/id/838DB18FA8EC26D87848FDEC1D50CFDC
  url = aws_eks_cluster.eks_cluster.identity[0].oidc[0].issuer
}
```

### Expected Behavior
The `url` should support using a HTTP proxy. Currently https://github.com/hashicorp/terraform-provider-tls/blob/0fe77c596df5141b7f44b20643f92b439baccf63/internal/provider/data_source_tls_certificate.go#L92 does not.

### Actual Behavior
Connection eventually times out (`connection timed out`)

### References

- Perhaps adjusting/replacing the dialer with something similar to https://github.com/mwitkow/go-http-dialer would add HTTP proxy support

detro:
Close #96 
Supersedes #113 

By introducing support for Proxy on the Provider level, we enable the `tls_certificate` data source (as well as future resources/data sources), to reach out endpoints via the optionally configured `proxy`.

The Proxy can be configured via a `url` (with optional `username` and `password), or `from_env`: this means that we leverage the excellent https://pkg.go.dev/net/http#ProxyFromEnvironment API to look for the environment variables and configure the proxy that way.

### DIFF
--- a/internal/provider/data_source_certificate.go
+++ b/internal/provider/data_source_certificate.go
@@ -189,7 +189,7 @@ func fetchPeerCertificatesViaHTTPS(targetURL *url.URL, shouldVerifyChain bool, c
 		},
 	}
 
-	// Fist attempting an HTTP HEAD: if it fails, ignore errors and move on
+	// First attempting an HTTP HEAD: if it fails, ignore errors and move on
 	resp, err := client.Head(targetURL.String())
 	if err == nil && resp.TLS != nil && len(resp.TLS.PeerCertificates) > 0 {
 		defer resp.Body.Close()


internal/provider/data_source_certificate.go
```diff
@@ -139,3 +160,33 @@ func dataSourceCertificateRead(d *schema.ResourceData, _ interface{}) error {
 
 	return nil
 }
+
+func fetchPeerCertificatesViaTLS(targetURL *url.URL, shouldVerifyChain bool) ([]*x509.Certificate, error) {
+	conn, err := tls.Dial("tcp", targetURL.Host, &tls.Config{
+		InsecureSkipVerify: !shouldVerifyChain,
+	})
+	if err != nil {
+		return nil, err
```
bflad, line 169: Nit: Might be good to return additional error context here

```suggestion
		return nil, fmt.Errorf("unable to open TLS connection: %w", err)
```

internal/provider/data_source_certificate.go
```diff
@@ -139,3 +160,33 @@ func dataSourceCertificateRead(d *schema.ResourceData, _ interface{}) error {
...
+func fetchPeerCertificatesViaHTTPS(targetURL *url.URL, shouldVerifyChain bool, config *providerConfig) ([]*x509.Certificate, error) {
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: !shouldVerifyChain,
+			},
+			Proxy: config.proxyForRequestFunc(),
+		},
+	}
+	resp, err := client.Get(targetURL.String())
```
bflad, line 185: Previously the data source would only open a TLS connection and not attempt to finish an actual HTTP call (in this new code the `GET` verb) or read the response body from the URL. A few questions:

* Can we avoid downloading the response body? e.g. should this execute a `HEAD` request or might a custom request handler function help here?
* Do we need to do anything to avoid new errors from switching to net/http?

bflad (commented):
Overall looking pretty good, just some things to consider around the new net/http and `from_env` handling.

internal/provider/provider.go
```diff
@@ -16,5 +24,99 @@ func New() *schema.Provider {
...
+						},
+						"from_env": {
+							Type:          schema.TypeBool,
+							Optional:      true,
+							Default:       false,
+							ConflictsWith: []string{"proxy.0.url", "proxy.0.username", "proxy.0.password"},
+							Description: "When `true` the provider will discover the proxy configuration from environment variables. " +
+								"This is based upon [`http.ProxyFromEnvironment`](https://pkg.go.dev/net/http#ProxyFromEnvironment) " +
+								"and it supports the same environment variables (default: `true`).",
+						},
```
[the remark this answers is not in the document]
bflad: For the first item, the HTTP server might need a way to reject connections except those from the proxy.

go.mod
```diff
@@ -3,6 +3,8 @@ module github.com/terraform-providers/terraform-provider-tls
 go 1.17
 
 require (
+	github.com/elazarl/goproxy v0.0.0-20220328115640-894aeddb713e
```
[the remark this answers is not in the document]
detro: Yeah, it's a bit annoying the way goproxy doesn't follow semver properly.
But in all my research it seemed like the best solution for what I needed to do.

Aside for 👍 -ing the issue, anything else I could do to improve this on our end?

internal/provider/data_source_certificate.go
detro: About using `HEAD` vs `GET` comes out of my experience that out there there are still HTTP servers that don't implement `HEAD` by default. In the past, where I had to do this kind of stuff, I'd even have logic to try `HEAD-then-GET`. I'm frankly unsure what's the best option here.

About the full HTTP call, I'm also unsure. It's out of this uncertainty that I introduced the support for `tls://` vs `https:/` schemes: my thinking was, until now we enforced `https:/` but actually never executed HTTP in the first place - just opened a socket.

So, I thought that adding `tls://` was going to be an easy way to suggest a "if you want to keep the previous behaviour...".

I'm frankly unsure what's the best strategy here. I'm tempted to suggest to make this PR a breaking change, because the only alternative I see is a `.mode` attribute that allows to chose the way the resource works. And I don't like it much.

internal/provider/provider.go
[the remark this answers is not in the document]
detro: 1. will add tests to use `t.Setenv` 👍 
2. Will fix the description
3. I'm not sure, my intention wasn't to make the `from_env` default: are you suggesting it should be made the default in a major release?
4. I can look into ways to orchestrate a "reject if not via proxy": I can try to play with headers injection more, but I have noticed that, for what our provider does, we never get to exchange http body - so requests never actually get to exchange headers. Will give it another pass.

internal/provider/data_source_certificate.go
bflad: Just to capture some of our out-of-band conversation (please correct me if I misspeak anywhere!):

1. We are going to try and implement this so the HTTP client is avoided, continuing to use a direct TLS connection instead, unless a HTTP proxy has been configured (currently via the provider `proxy` configuration block)
2. Since IANA does not have a URI scheme for strictly a TCP network socket connection, we can use a synthetic `tls://` scheme (or potentially use _no_ scheme).
3. When using a HTTP proxied client, it will first attempt a `HEAD` call, then fallback to a `GET` call in case of a proxy that does not support `HEAD` and blocks the subsequent HTTP connection.

Not discussed, but thinking out loud: I'm presuming we can capture 405 Method Not Allowed or 501 Not Implemented HTTP status codes for the HTTP proxied client fallback. Another thing I'm not sure about is whether we will need special 407 Proxy Authentication Required handling anywhere.

internal/provider/provider.go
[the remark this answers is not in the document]
bflad: > my intention wasn't to make the from_env default: are you suggesting it should be made the default in a major release?

I think we should certainly consider this for the next major version. 😄  Many pieces of software handling HTTP calls will conventionally support the `NO_PROXY`/`HTTP_PROXY`/`HTTPS_PROXY` environment variables without additional configuration. github.com/hashicorp/go-cleanhttp default clients for example, implement net/http.ProxyFromEnvironment, which I believe is used by the majority of HTTP handling within Terraform CLI itself.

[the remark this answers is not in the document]
detro: So, the last point you make is great, but it kinda goes against the other thing we decided: to not use an HTTP client unless a Proxy is configured.

Using `ProxyFromEnvironment` by default would also change the current behaviour.

So, I think now I follow better what you head in mind earlier when you suggested:
> Should we note that this will be enabled by default in the next major version and add a code comment so that gets done?

In the light of this, I think I'll do the following:
* by default, if no `proxy` config is provided, we keep using `tls.Client`
* if a proxy is configured, then we switch the logic to use `http.Client`
* `from_env` is defaulted to `false`

Then, I'll create a follow up issue for breaking changes that we want to launch in 4.x (similar to [this one](https://github.com/hashicorp/terraform-provider-tls/issues/174)). In this issue I'll include:
* `from_env` becomes defaulted to `true`
* the URL scheme defines the client we will use: this means `https://` will always use the `http.Client`, while `tls://` will always use `tls.Client`

Does this strategy sound good? Am I missing something?

internal/provider/data_source_certificate.go
detro: Actually, while I want to handle `HEAD` then `GET`, I think we can't give lots of weight to the actual HTTP response: the TLS negotiation happens much earlier then any HTTP Protocol byte has moved.

So, I think it makes sense to handle error response from an `http.Get()` call, but if the `reps.TLS.PeerCertificates` are there, this provider shouldn't do much more than process those.

I'll show you what I mean in the next commits I push.

detro:
@bflad I pushed the changes as discussed above, but I need to call out a few things:

* no point in checking status code when what our code is trying to do is grab the certificates exchanged as part of the TLS handshake: we can rely in our logic exclusively on the `err`s returned
* if you look at the set of tests, they should prove all the error scenarios you mentioned (authentication required + unreachable proxy)
* I have created https://github.com/hashicorp/terraform-provider-tls/issues/183 as the follow up (things to change in the major release)

I have started accumulating the "major release stuff" in a 4.0.0 milestone.

internal/provider/data_source_certificate.go
```diff
@@ -139,3 +166,45 @@ func dataSourceCertificateRead(d *schema.ResourceData, _ interface{}) error {
...
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: !shouldVerifyChain,
+			},
+			Proxy: config.proxyForRequestFunc(),
+		},
+	}
+
+	// Fist attempting an HTTP HEAD: if it fails, ignore errors and move on
```
bflad, line 192: Typo 😎 

```suggestion
	// First attempting an HTTP HEAD: if it fails, ignore errors and move on
```

internal/provider/data_source_certificate.go
```diff
@@ -104,32 +112,51 @@ func dataSourceCertificate() *schema.Resource {
...
+	var peerCerts []*x509.Certificate
+	switch targetURL.Scheme {
+	case HTTPSScheme.String():
+		if targetURL.Port() == "" {
+			targetURL.Host += ":443"
+		}
+
+		// TODO remove this branch and default to use `fetchPeerCertificatesViaHTTPS`
+		//   as part of https://github.com/hashicorp/terraform-provider-tls/issues/183
+		if config.isProxyConfigured() {
```
bflad, line 136: I still think the next major version should still perform this type of check and use the TLS client whenever possible for simplicity -- the check could see if `http.ProxyFromEnvironment` would actually use a proxy for the request by checking for `nil`.

bflad (approved):
Looks good to me 🚀

internal/provider/data_source_certificate.go
detro: The reason I insist on using an actual HTTP client, is that we were and are requiring a `scheme` for the `url` argument. The fact that practitioners have to write `https://` is, in my opinion, an implicit implication we are making.

The original implementation simply didn't respect that. It would have made more sense if there was no scheme, and then it only (and exclusively) used `tls.Client`. Though that would have prevented even talking about supporting HTTP Proxies.

PR #184, opened 2022-04-07
Relaxing validation of `allowed_uses` in `tls_locally_signed_cert` and `tls_self_signed_cert`

detro:
As part of [milestone 3.2.0](https://github.com/hashicorp/terraform-provider-tls/milestone/1), more stringent validation was introduced in many arguments fields. One of which is the argument `allowed_uses`, used by the resources:

* `tls_locally_signed_cert`
* `tls_self_signed_cert`

Given the nature of the argument, values provided that were not known to the Provider code, were silently ignored: the practitioner would get a certificate without one or more of the [Key Usages](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3) they expected, but the provider would not report it.

Hence, the introduction of the validation.

Unfortunately this means that existing terraform configurations started failing with the update to 3.2.0. 

So, in the interest of not introducing regressions, this PR alters the validation logic slightly: if practitioner uses a non acceptable value for `allowed_uses`, instead of erroring, **the provider will raise a warning and ignore the specific erroneous argument**.

### DIFF
--- a/internal/provider/common_cert.go
+++ b/internal/provider/common_cert.go
@@ -270,7 +270,7 @@ func setCertificateCommonSchema(s map[string]*schema.Schema) {
 		ForceNew: true,
 		Elem: &schema.Schema{
 			Type:         schema.TypeString,
-			ValidateFunc: validation.StringInSlice(supportedKeyUsages(), false),
+			ValidateFunc: StringInSliceOrWarn(supportedKeyUsages(), false),
 		},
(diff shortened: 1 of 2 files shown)


bflad (approved):
This is a great quick fix, nice! 🚀  Should we consider hard validation again for 4.0? 😄

bflad:
Andddd you're already on it, rock on. https://github.com/hashicorp/terraform-provider-tls/issues/185

PR #139, opened 2021-12-22
Support local tls certificate file as input for tls_certificate data source 

dttung2905:
Hi team,

I would like to propose a new feature to support local tls file as an input for `tls_certificate` data source.

Example:

For local file
```
data "tls_certificate" "test" {
   input  =  file("testdata/tls_certs/public.pem")
   type   = "local"
```
For remote file (sitting behind and HTTPS endpoint)
```
data "tls_certificate" "test" {
 input = "https://your-target-website.com"
 type  = "remote"

```

This will serve as an extension for  #62  solve #4  where the author wants to read from file `${file('path/to/cert.pem')}` for example

PR #189, opened 2022-04-12
support parsing certificate content #4

### LINKED ISSUE #79: Consistent ID across multiple planning
### Terraform Version

```
terraform -v
Terraform v0.13.2
+ provider registry.terraform.io/hashicorp/tls v2.2.0
```
### Affected Resource(s)
- tls_certificate

```hcl
data "tls_certificate" "oidc_idp" {
  url = "https://oidc.eks.sa-east-1.amazonaws.com/id/<REDACTED>"
}
```

### Expected Behavior
Consistent `id` across consecutive plans.

### Actual Behavior
Resource `id` changes every plan even if certificates are the same.

```
...
      ~ id           = "2020-09-14 19:34:14.498463939 +0000 UTC" -> "2020-09-14 19:34:14.722109742 +0000 UTC"
...
```

### Steps to Reproduce
1. `terraform plan .`
2. `terraform apply .`
3. `terraform plan .`

### Important Factoids
Obviously, this is the behavior given by [this line](https://github.com/hashicorp/terraform-provider-tls/blob/f1b91f494149cc797fd710691c66fcf8a73c9469/tls/data_source_tls_certificate.go#L109).
Is this intentional? Could it be something stable, like the certificate's serial number or sha1 fingerprint?

iwarapter:
This allows directly parsing of certificates, i.e. loaded by files etc.

Closes #139

### Update by provider maintainers

As part of work on finalizing this PR, an existing issue of this provider surfaced (see https://github.com/hashicorp/terraform-provider-tls/pull/189#issuecomment-1117549152). So, we added a fix for this, in addition to the main focus of this PR.

Closes #79


RalphBragg:
This would be a great feature!

internal/provider/data_source_certificate.go
```diff
@@ -24,18 +25,26 @@ func dataSourceCertificate() *schema.Resource {
...
 					fmt.Sprintf("Accepted schemes are: `%s`. ", strings.Join(SupportedURLSchemesStr(), "`, `")) +
 					"For scheme `https://` it will use the HTTP protocol and apply the `proxy` configuration " +
 					"of the provider, if set. For scheme `tls://` it will instead use a secure TCP socket.",
 				ValidateDiagFunc: validation.ToDiagFunc(validation.IsURLWithScheme(SupportedURLSchemesStr())),
+				ConflictsWith:    []string{"content"},
+			},
+			"content": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "The content of the PEM encoded certificate",
```
detro, line 39: Like for other resources and data sources, please link in the descriptions the RFCs describing the format of PEM expected here ([examples](https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key#private_key_pem)).

Also, please terminate the description with a `.`.

detro: Also, as `encoded` is a loaded term, please use the `PEM format` or `PEM formatted` expression here, in consistency with the other resources.

internal/provider/data_source_certificate.go
```diff
@@ -115,51 +124,69 @@ func dataSourceCertificate() *schema.Resource {
...
-	switch targetURL.Scheme {
-	case HTTPSScheme.String():
-		if targetURL.Port() == "" {
-			targetURL.Host += ":443"
+	if v, ok := d.GetOk("content"); ok {
+		block, _ := pem.Decode([]byte(v.(string)))
+		if block == nil {
+			return fmt.Errorf("failed to decode pem content")
+		}
+		if block.Type != "CERTIFICATE" {
```
detro, line 132: Please use the `PEMPreamble` const available in `types.go`.

detro: You should also use `PEMBlockToPEMPreamble` to extract the correct `PEMPreamble`.

internal/provider/data_source_certificate.go
```diff
@@ -115,51 +124,69 @@ func dataSourceCertificate() *schema.Resource {
...
+			return fmt.Errorf("pem must be of type 'CERTIFICATE'")
```
detro, line 133: `PEM` over `pem`

internal/provider/data_source_certificate_test.go
```diff
@@ -8,6 +8,53 @@ import (
...
+				ExpectError: regexp.MustCompile("failed to decode pem content"),
+			},
+			{
+				Config: `
+					data "tls_certificate" "test" {
+					  content = file("testdata/tls_certs/private.pem")
+					}
+				`,
+				ExpectError: regexp.MustCompile("pem must be of type 'CERTIFICATE'"),
+			},
```
detro, line 53: Could you also add negative cases to prove that it's not possible to use both `url` and `content` at the same time?

Additionally, it would be nicer if the tests were split a bit more, at least to separate "positive" and "negative" cases.

detro (changes requested):
Thank you so much for this @iwarapter.
I have a few requests for you before we proceed, additional to the comments you will find inline.

1. please use `make generate` to update the documentation
2. please create a new `CHANGELOG` entry for a "minor version" bump (I believe this will be `3.4.0` as we stand?)
3. please indicate in the description of the PR that this `Closes #139` as well

Looking forward to the refinements so we can release this new feature.

Again, thank you for your contribution!

detro:
@iwarapter ping :)

I have a 4.x release lined up (there are a few breaking changes that need to go in). If we sort this out in time, we could add it to that train.

Do you have time to address the issues I highlighted?

iwarapter:
@detro I have addressed all the changes you requested the `PEMPreamble` doesn't actually add any benefit just extra code 🤷 ?

examples/data-sources/tls_certificate/data-source.tf
```diff
@@ -11,3 +11,7 @@ resource "aws_iam_openid_connect_provider" "example" {
   thumbprint_list = [data.tls_certificate.example.certificates[0].sha1_fingerprint]
   url             = aws_eks_cluster.example.identity[0].oidc[0].issuer
 }
+
+data "tls_certificate" "example_content" {
+  content = file("example.pem")
+}
```
detro, line 17: I'm not a fan of having 2 separate examples in the same example `.tf`, but if you are going this route, please add at least 1 line of comment for each to describe.

internal/provider/data_source_certificate.go
```diff
@@ -115,51 +124,73 @@ func dataSourceCertificate() *schema.Resource {
...
+	if v, ok := d.GetOk("content"); ok {
+		block, _ := pem.Decode([]byte(v.(string)))
+		if block == nil {
+			return fmt.Errorf("failed to decode pem content")
+		}
+		preamble, err := PEMBlockToPEMPreamble(block)
+		if err != nil {
+			return err
+		}
+		if preamble != "CERTIFICATE" {
```
detro, line 136: please use the `PreambleCertificate` `const` here: it's about the consistency and the single-place-definition of key constants like this one.

detro (commented):
A few more tweaks and we are in business.

examples/data-sources/tls_certificate/data-source.tf
detro: Alternatively, if you want to split in 2 files, you can do that and then alter the template https://github.com/hashicorp/terraform-provider-tls/blob/main/templates/data-sources/certificate.md.tmpl to include both.

I had to do something similar for the examples of configuration of the Provider: it does allow to add a bit of verbal explanation between examples that it might be nice.

iwarapter: @detro did this already :)

iwarapter:
hi @detro sorry looks like i forgot to run the `make generate` whilst i split the example files out.

detro (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 3.4.0
```
detro, line 1:
```suggestion
## 3.4.0 (UNRELEASED)
```

detro:
I think there might be something that need fixing, in regards to the ID generation: https://github.com/hashicorp/terraform-provider-tls/runs/6175464609?check_suite_focus=true#step:5:65

iwarapter:
hmm this works on all TF versions bar 13.x which feels more like a TF bug that a provider bug? The ID generation is the same from before my change. But I don't know enough of the TF core to debug it.

detro:
Also, could you rebase this too?

detro:
I just rebased

detro:
Yep, this seems to sort things out.

Issue #197, opened 2022-04-29
Unable to Import "ED25519" key into resource "aws_key_pair"


jseiser:
### Terraform CLI and Provider Versions

Terraform Version: 1.1.8

Initializing provider plugins...
- Finding latest version of hashicorp/random...
- Finding hashicorp/aws versions matching ">= 3.0.0, >= 3.63.0, >= 3.67.0, >= 3.72.0, ~> 4.0"...
- Finding latest version of hashicorp/tls...
- Installing hashicorp/random v3.1.3...
- Installed hashicorp/random v3.1.3 (signed by HashiCorp)
- Installing hashicorp/aws v4.12.0...
- Installed hashicorp/aws v4.12.0 (signed by HashiCorp)
- Installing hashicorp/tls v3.3.0...
- Installed hashicorp/tls v3.3.0 (signed by HashiCorp)

### Terraform Configuration

```terraform
resource "tls_private_key" "key" {
  algorithm = "ED25519"
}

resource "aws_key_pair" "generated_key" {
  key_name   = format("%s-%s", var.vpc_id_tag, var.platform_env)
  public_key = tls_private_key.key.public_key_openssh
  tags       = local.tags
}
```

### Expected Behavior

The new ED25519 key should have been imported into AWS.

### Actual Behavior

Error: error importing EC2 Key Pair (JLS-DEV): InvalidKey.Format: Key is not in valid OpenSSH public key format
	status code: 400, request id: e9d5cfb0-d0c5-4ab1-90c6-4476526cb2ca

  with aws_key_pair.generated_key,
  on ssh.tf line 5, in resource "aws_key_pair" "generated_key":
   5: resource "aws_key_pair" "generated_key" {

### Steps to Reproduce

1. `terraform apply`

### How much impact is this issue causing?

Medium

### Additional Information

The above code, is what we use currently, to generate keys for the dev environments.  but instead of ED225519, we use RSA.

The docs show `public_key_openssh` should still be valid, and the AWS docs show this type of key is supported and that my provider version is new enough to support it.

* https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key#public_key_openssh
* https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html
* https://github.com/hashicorp/terraform-provider-tls/pull/151

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

detro:
[...]

Issue #163, opened 2022-03-09
Remove "\n" at the end of public_key_openssh in tls_private_key resource


ujjwal-ibm:
### Terraform CLI and Provider Versions

Terraform v1.0.9

### Terraform Configuration

```terraform
resource "tls_private_key" "insecure" {
  algorithm = "RSA"
  rsa_bits  = 4096
}
```

### Expected Behavior

public_key_openssh attribute should be generated without new line character

### Actual Behavior

public_key_openssh attribute is generated with new line character

           ` "public_key_openssh": "ssh-rsa ......==\n",`

### Steps to Reproduce

1. `terraform apply`

### How much impact is this issue causing?

High

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

detro:
Hello @ujjwal-ibm, thanks for opening this issue.

So, I did a bit of research, and it turns out that it's an established convention to generate OpenSSH Public Key in a `~/.ssh/authorized_keys` friendly format. That means that each entry is ready to be appended to `authorized_keys` as is. And the newline works as a "termination" for the key.

I can't pin point to an RFC for this unfortunately, but suggest a bit of googling around.

Additional to this, I have looked at how exactly that newline get appended: turns out this is baked into [`MarshalAuthorizedKey(key PublicKey)`](https://github.com/golang/crypto/blob/master/ssh/keys.go#L288) in `x/crypto/ssh` itself. This is the official cryptography library for Golang, so I tend to rely on the fact that who wrote this knows more than I do on what's right for those formats and conventions.

Lastly, even if we wanted to make a change like this, it would most definitely cause issues for existing configuration that relies and handles the field as carrying that trailing `\n`: removing it will most certainly cause issues for those codebases, and it wouldn't gain us much (getting rid of it is a `trim()` call away in pretty much any language).

I hope this clarifies why we will not be changing this.

PR #198, opened 2022-04-29
Updating doc for PEM fields that underlying Golang libraries append a `\n` to

detro:
Closes: #197
Additional context: #163
Additional context: #9
Relates to: https://github.com/hashicorp/terraform-provider-aws/issues/24025

Adding a suggestion to use `trimspace()` in case the generated PEM causes issues.

One thing I'm not sure about: does it make sense to mention the underlying libraries that "cause" this, or not.


bflad (approved):
Looks good to me 🚀

bendbennett (approved):

PR #203, opened 2022-05-04
public_key_fingerprint_x509_sha256

### LINKED ISSUE #202: Support x509/DER/SHA256 public key fingerprint

Terraform v1.1.9
on darwin_arm64

When trying to configure an OIDC provider, I needed a [particular checksum ](https://github.com/aws/amazon-eks-pod-identity-webhook/blob/0d254eee1537e0745679252ca60f020fa1a461f0/hack/self-hosted/main.go#L19-L33) to be supported; namely the SHA256 hash of the DER-formatted public key.

### Proposal

I propose that a new attribute, `public_key_fingerprint_x509_sha256` is added to the `tls_private_key` resource and the `tls_public_key` data source.

The hash algorithm seems to use standard encryption methods and Go libraries and hence seems to align with the design guidelines. However, it may cause property bloat to the `tls_private_key` resource and `tls_public_key` data source.

Medium

### Additional Information

I have proposed the change in this provider, rather than writing a separate one, as it already has access to the key material (and indeed uses the x509 package already).

### Code of Conduct

alanraison:
Resolves #202

I am a little concerned with the long name of this property, though it fits in with the other property names, and also whether the name is accurate.

I believe this is a good location for this fingerprint, rather than in a separate provider, since the key material is already available.


detro:
Hello and thanks for providing this.

Before I proceed further, I have to ask: why is this property https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key#public_key_fingerprint_sha256 not enough for your purpose?

alanraison:
Hi there @detro. Unfortunately the ssh-format hash is not the same as used by kubernetes to calculate the Key Id during OIDC authentication. I am trying to use a tls_private_key to populate a jwks key set for my cluster, so the hash algorithm has to match.

I have to agree that it doesn't feel quite right to have arbitrary hash methods in this resource, and I can't find any evidence that the kubernetes method is at all standard (other than by using similarly available go crypto functions), so if you have any suggestions about where else this could live I'd be happy to consider it.

thefirstofthe300:
One potential way to address the issue here from what I can tell would be to return the base64 encoded value of the DER encoding. From there, Terraform would be able to base64 decode the value, sha256sum it, and base64 encode it back using built-in functions.

Providing only the base64 encoded DER value is arguably more flexible for future use cases.

For context, I'm trying to solve the same problem as @alanraison.

PR #207, opened 2022-05-11
CI to build against both 1.17 and 1.18 for future-proofing

detro:
Context: https://github.com/hashicorp/terraform-provider-tls/issues/188#issuecomment-1123404375


bendbennett (approved):
LGTM

PR #209, opened 2022-05-11
Allow empty or absent `subject` block

### LINKED ISSUE #61: subject block required, all fields optional, but cannot pass empty block
### Terraform Version
```
Terraform v0.12.10
+ provider.tls v2.1.1
```

### Affected Resource(s)
- `tls_cert_request`
- `tls_self_signed_cert`

- fails:
  ```hcl
  resource "tls_private_key" "this" {
    algorithm = "RSA"
  }
  resource "tls_self_signed_cert" "this" {
    early_renewal_hours   = 720
    key_algorithm         = "RSA"
    private_key_pem       = tls_private_key.this.private_key_pem
    validity_period_hours = 2160
  }
  ```
- fails:
  ```hcl
  resource "tls_private_key" "this" {
    algorithm = "RSA"
  }
  resource "tls_self_signed_cert" "this" {
    early_renewal_hours   = 720
    key_algorithm         = "RSA"
    private_key_pem       = tls_private_key.this.private_key_pem
    validity_period_hours = 2160
    subject {}
  }
  ```
- fails:
  ```hcl
  resource "tls_private_key" "this" {
    algorithm = "RSA"
  }
  resource "tls_self_signed_cert" "this" {
    early_renewal_hours   = 720
    key_algorithm         = "RSA"
    private_key_pem       = tls_private_key.this.private_key_pem
    validity_period_hours = 2160
    subject {
      common_name = "" # <--- note the empty string
    }
  }
  ```
- succeeds:
  ```hcl
  resource "tls_private_key" "this" {
    algorithm = "RSA"
  }
  resource "tls_self_signed_cert" "this" {
    early_renewal_hours   = 720
    key_algorithm         = "RSA"
    private_key_pem       = tls_private_key.this.private_key_pem
    validity_period_hours = 2160
    subject {
      common_name = " " # <--- note the space... this is the kludge to get it to work
    }
  }
  ```

### Expected Behavior
Terraform generates the `tls_self_signed_cert` resource with no subject field on the certificate, and Terraform demands no `subject` block at all.

### Actual Behavior
Fails with this error when no subject block:
https://github.com/terraform-providers/terraform-provider-tls/blob/afd9e9546f57b662e375ef4146c0209e1b75c6f6/tls/resource_cert_request.go#L94

Fails with this error when an empty subject block:
https://github.com/terraform-providers/terraform-provider-tls/blob/afd9e9546f57b662e375ef4146c0209e1b75c6f6/tls/resource_cert_request.go#L98

Fails with this (same) error when subject block has single field explicitly set to an empty string `common_name =  ""`:
https://github.com/terraform-providers/terraform-provider-tls/blob/afd9e9546f57b662e375ef4146c0209e1b75c6f6/tls/resource_cert_request.go#L98

### Steps to Reproduce
1. `terraform apply` with the first set of configs (labelled "fails") above.

### Important Factoids
I believe this was introduced in https://github.com/terraform-providers/terraform-provider-tls/pull/7. That issue is addressing a (valid) user behavior of providing no subject at all (https://github.com/terraform-providers/terraform-provider-tls/issues/1) which had been causing a panic.

I think this all stems from type coercion:
https://github.com/terraform-providers/terraform-provider-tls/blob/afd9e9546f57b662e375ef4146c0209e1b75c6f6/tls/resource_cert_request.go#L96

... which is only necessary because `subject` is (unnecessarily) marked required:
https://github.com/terraform-providers/terraform-provider-tls/blob/afd9e9546f57b662e375ef4146c0209e1b75c6f6/tls/resource_cert_request.go#L71-L76

Of note, the code which parses the subject block's fields treats all of them as optional:
https://github.com/terraform-providers/terraform-provider-tls/blob/afd9e9546f57b662e375ef4146c0209e1b75c6f6/tls/provider.go#L35-L70

### References
- https://tools.ietf.org/html/rfc5280#section-4.1.2.6 implies that an empty subject field is valid:
  >Where it is non-empty, the subject field MUST contain an X.500
   distinguished name (DN)

detro:
Closes #61 

RFC that motivates fixing this bug: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.6

This affects 2 resources:

* `tls_cert_request`
* `tls_self_signed_cert`

### Caveat
In accordance with RFC 5280, the creation of the certificate will 
fail, if the desired certificate is a CA _and_ it has an empty Subject (i.e. no Distinguished Name is provided). Specifically, this part of the RFC:

> If the subject is a CA (e.g., the basic constraints extension, as discussed in [Section 4.2.1.9](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9), is present and the value of cA is TRUE), then the subject field MUST be populated with a non-empty distinguished name matching the contents of the issuer field in all certificates issued by the subject CA.

### Small side improvement

Something that is possible to do with `tls_locally_signed_cert`, but it's probably undesirable, is to create a certificate using as CA a certificate that, in facts, is not one.

As this if functionally legit, now we will just raise a warning every time a certificate with this "quirk" gets created with this resource.

```
╷
│ Warning: CA is not a CA
│
│   with tls_locally_signed_cert.test,
│   on main.tf line 24, in resource "tls_locally_signed_cert" "test":
│   24: resource "tls_locally_signed_cert" "test" {
│
│ Certificate provided as Authority does not appear to be as much: the resulting certificate might fail
│ validations
╵
```

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -8,6 +8,7 @@ install: build
 
 # See https://golangci-lint.run/
 lint:
+	golangci-lint cache clean
 	golangci-lint run
 
 generate:
--- a/internal/provider/common_cert.go
+++ b/internal/provider/common_cert.go
@@ -456,11 +456,9 @@ func updateCertificate(_ context.Context, _ *schema.ResourceData, _ interface{})
 	return nil
 }
 
-// getSubjectDistinguishedNames it looks for the "subject" attribute: if found and non-empty, it returns
-// a *pkix.Name useful for creating x509.Certificate or x509.CertificateRequest.
-func getSubjectDistinguishedNames(d *schema.ResourceData) *pkix.Name {
-	subjectList := d.Get("subject").([]interface{})
-
+// createSubjectDistinguishedNames return a *pkix.Name (i.e. a "Certificate Subject") if the non-empty.
+// This used for creating x509.Certificate or x509.CertificateRequest.
+func createSubjectDistinguishedNames(subjectList []interface{}) *pkix.Name {
 	if len(subjectList) == 0 {
 		// No subject block was provided
 		return nil
--- a/internal/provider/common_cert.go
+++ b/internal/provider/common_cert.go
@@ -19,6 +19,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
@@ -354,6 +355,12 @@ func createCertificate(d *schema.ResourceData, template, parent *x509.Certificat
 	}
 
 	if d.Get("is_ca_certificate").(bool) {
+		// NOTE: if the Certificate we are trying to create is a Certificate Authority,
+		// then https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.6 requires
+		// the `.Subject` to contain at least 1 Distinguished Name.
+		if cmp.Equal(template.Subject, pkix.Name{}) {
+			return diag.Errorf("Certificate Subject must contain at least one Distinguished Name when creating Certificate Authority (CA)")
+		}
 		template.IsCA = true
 
 		template.SubjectKeyId, err = generateSubjectKeyID(pub)
--- a/internal/provider/common_cert.go
+++ b/internal/provider/common_cert.go
@@ -164,8 +164,9 @@ func setCertificateSubjectSchema(s map[string]*schema.Schema) {
 
 	s["subject"] = &schema.Schema{
 		Type:     schema.TypeList,
-		Required: true,
+		Optional: true,
 		ForceNew: true,
+		MaxItems: 1,
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
 				"organization": {
--- a/internal/provider/common_cert.go
+++ b/internal/provider/common_cert.go
@@ -451,18 +451,18 @@ func updateCertificate(_ context.Context, _ *schema.ResourceData, _ interface{})
 
 // getSubjectDistinguishedNames it looks for the "subject" attribute: if found and non-empty, it returns
 // a *pkix.Name useful for creating x509.Certificate or x509.CertificateRequest.
-func getSubjectDistinguishedNames(d *schema.ResourceData) (*pkix.Name, error) {
+func getSubjectDistinguishedNames(d *schema.ResourceData) *pkix.Name {
 	subjectList := d.Get("subject").([]interface{})
 
 	if len(subjectList) == 0 {
 		// No subject block was provided
-		return nil, nil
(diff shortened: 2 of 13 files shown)


bookshelfdave (approved):
lgtm! 👍

internal/provider/common_cert.go
```diff
@@ -448,39 +449,52 @@ func updateCertificate(_ context.Context, _ *schema.ResourceData, _ interface{})
 	return nil
 }
 
-// distinguishedNamesFromSubjectAttributes it takes a map subject attributes and
-// converts it to a pkix.Name (X.509 distinguished names).
-func distinguishedNamesFromSubjectAttributes(nameMap map[string]interface{}) *pkix.Name {
+// getSubjectDistinguishedNames it looks for the "subject" attribute: if found and non-empty, it returns
+// a *pkix.Name useful for creating x509.Certificate or x509.CertificateRequest.
+func getSubjectDistinguishedNames(d *schema.ResourceData) *pkix.Name {
```
bflad, line 454: Totally a small nit, but something I've found super helpful for potentially adding unit testing (and maybe helping with the framework migration) is creating functions which use the attributes value type (in this case a `[]interface{}` parameter) where the calling code does the necessary work to get data out of `ResourceData` (in this case `getSubjectDistinguishedNames(d.Get("subject").([]interface{}))`).

internal/provider/resource_self_signed_cert_test.go
```diff
@@ -508,3 +531,114 @@ func TestAccResourceSelfSignedCert_FromRSAPrivateKeyResource(t *testing.T) {
...
+						]
+						ip_addresses = [
+							"127.0.0.2",
+						]
+						uris = [
+							"disney://pippo.pluto.paperino/minnie",
+						]
+					}
+				`,
+				Check: r.ComposeTestCheckFunc(
```
bflad, line 568: Small nits:

- It might be cool to switch this to `ComposeAggregateTestCheckFunc()` similar to other so all check failures are shown at once, rather than only the first.
- Would it be worth having a `testCheckPEMCertificateNoSubject()` helper just to double check things?

bflad (approved):
Looks good to me, yay for being more RFC compliant. 🚀  Some small nits for consideration, nothing blocking from me. 😄

bendbennett (approved):
LGTM

bendbennett (approved):
LGTM

GNUmakefile
```diff
@@ -8,6 +8,7 @@ install: build
 
 # See https://golangci-lint.run/
 lint:
+	golangci-lint cache clean
```
bflad, line 11: Interesting, I've never needed to run this before. Is there something going on that requires it?

detro: I haven't experienced this before, but I have had a situation with the `unparam` linter where it wouldn't pick up an issue until I cleaned the cache.

I might be something that can only happen locally, as the runners on the GH CI are always executed against a fresh environment.

I'm not sure it's necessary, but it did seem to make a difference in spotting [this](https://github.com/hashicorp/terraform-provider-tls/pull/209/files#diff-cba759d368325cfd69f2c0b0bc003b2d4bee940fd86255e00549060ef711e175R48). 🤷

PR #224, opened 2022-06-27
Making Provider's `proxy.from_env` default to `true`

### LINKED ISSUE #183: Make provider `proxy.from_env = true` default + enforce "scheme determines client" in `tls_certificate` data source

* cli >= 0.12
* provider >= 3.2.1

As part of the work for #96 (#179) we have identified a couple of places where the overall design and behaviour of the provider around the `tls_certificate` data source can be improved.

* the provider should pick up Proxy configuration from env variables by default, like for many other providers (including the [AWS Provider](https://github.com/hashicorp/terraform-provider-aws/blob/5aad89743faae647682c752030cade7685a78733/internal/provider/provider.go#L244))
* using the argument `url = "https://..."` should imply that underneath you are using an actual HTTP client

Previous thread about this: https://github.com/hashicorp/terraform-provider-tls/pull/179#discussion_r844039017

### Proposal

As part of the 4.x major release of this provider, we should:

* make the provider discover Proxy configuration from env vars by default (i.e. `proxy.from_env = true` by default, instead of the current `false`)
* enforce `https:// -> http.Client`, `tls:// -> tls.Client`

Low

### Additional Information

_No response_

### Code of Conduct

detro:
Closes #183


bendbennett (approved):
LGTM

bflad (approved):
Looks good to me 🚀 (minus the acceptance test failures)

detro:
Yeah, I'm trying to figure out what's going on with the fetching of proxy config from env.

It seems not be working, now that I have added an extra step to count incoming connections.

PR #225, opened 2022-06-27
Adding PKCS#8 PEM attribute to `tls_private_key`

### LINKED ISSUE #210: Expose `tls_private_key` PEM also in PKCS#8 (RFC 5208) format

0.12

In the current implementation of this provider, it's possible to provide Private Key PEMs in [PKCS#8 format](https://datatracker.ietf.org/doc/html/rfc5208) to certificate resources and data sources. Yet, is not possible to use `tls_private_key` to get the key in PEM PKCS#8 format.

To be exact, we _can_ get it only for `tls_private_key` created with `algorithm = ED25519`, as that's the only format Go crypto libraries support for it.

### Proposal

Add a `.private_key_pem_pkcs8` field to `tls_private_key` resource.

Low

### Additional Information

This missing feature was spotted [here](https://github.com/hashicorp/terraform-provider-tls/pull/94#issuecomment-1124063879).

Related #91 
Related #55

### Code of Conduct

detro:
Closes #210


bendbennett (approved):
LGTM

detro:
LOL - I was looking for the new field in the version 3.4.0 :P (hence the failure)

internal/provider/resource_private_key.go
```diff
@@ -290,3 +268,88 @@ func (r *privateKeyResource) Delete(ctx context.Context, _ tfsdk.DeleteResourceR
...
+
+func (r *privateKeyResource) UpgradeState(ctx context.Context) map[int64]tfsdk.ResourceStateUpgrader {
+	schema, _ := (&privateKeyResourceType{}).GetSchema(ctx)
+	return map[int64]tfsdk.ResourceStateUpgrader{
+		// Upgrading schema v0 -> v1 will add:
+		// * `private_key_openssh` (introduced in v3.2.0)
+		// * `public_key_fingerprint_sha256` (introduced in v3.2.0)
+		// * `private_key_pkcs8`   (introduced in v4.0.0)
+		0: {
+			PriorSchema: &schema,
```
bendbennett, line 280: `PriorSchema` appears to be using a schema with `Version: 1`. Should the version for the `PriorSchema` match the map key (i.e., `0`)?

bendbennett (approved):
LGTM 
Just a minor comment about the schema version being used.

internal/provider/resource_private_key.go
detro: Great catch.
I have left a long explanation in the code now:
```
// NOTE: why are we using a Schema v1 to unmarshal configuration with Schema v0?
// This is possible because the way the unmarshalling works, is that if a field
// is found in the RawSchema, is then set on the State by looking for a compatible
// field in the Schema.
//
// In other words, fields that are not present in the RawSchema, are simply ignored:
// the equivalency of RawSchema and Schema is not bidirectional.
//
// This works fine _as long as_ there are no renaming or retractions from the Schema.
// If/when that happens, and a new Schema version is released, then a dedicated
// schema will be necessary.
```

PR #233, opened 2022-07-11
Tweak acceptance test against terraform.io

detro:
* Don't change fingerprint of certificate of https://terraform.io: it's auto-generated and the fingerprint will change frequently
* Tweaked logging of the provider, to avoid printing out the proxy password
* Updated dependencies


bendbennett (approved):
LGTM

PR #235, opened 2022-07-11
Pre-4.x tweaks

detro:
Follow up from #225

This PRs touches upon a few PKCS#8 related issues we wanted to iron out before the 4.x release:

* Rename new attribute to private_key_pem_pkcs8
* Add mention of [RFC-5208](https://datatracker.ietf.org/doc/html/rfc5208) to the DESIGN.md
* Ensure each *_cert resource has a test to verify it can handle the new PKCS#8 PEM format
* Fix proxy configuration documentation to use nested attribute syntax


bendbennett (approved):
LGTM

PR #246, opened 2022-07-25
BUGFIX: use `ListType` of `ObjectType` for `Computed` attributes instead of `Blocks`

### LINKED ISSUE #244: data.tls_certificate.this.certificates is empty list of object

`terraform` 1.2.5
`hashicorp/tls` 4.0.0
`hashicorp/aws` 4.23.0

### Terraform Configuration

```terraform
data "tls_certificate" "this" {
  url = aws_eks_cluster.this.identity[0].oidc[0].issuer
}

resource "aws_iam_openid_connect_provider" "this" {
  client_id_list  = ["sts.amazonaws.com"]
  thumbprint_list = [data.tls_certificate.this.certificates[0].sha1_fingerprint]
  url             = aws_eks_cluster.this.identity[0].oidc[0].issuer
}
```

### Expected Behavior

`terraform plan` should work normally

### Actual Behavior

`terraform plan` fails with the following error
```
╷
│ Error: Invalid index
│
│   on ../../../eksv2/cluster/oidcp.tf line 7, in resource "aws_iam_openid_connect_provider" "this":
│    7:   thumbprint_list = [data.tls_certificate.this.certificates[0].sha1_fingerprint]
│     ├────────────────
│     │ data.tls_certificate.this.certificates is empty list of object
│
│ The given key does not identify an element in this collection value: the collection has no elements.
```

### Steps to Reproduce

Add

```hcl
data "tls_certificate" "this" {
  url = aws_eks_cluster.this.identity[0].oidc[0].issuer
}

resource "aws_iam_openid_connect_provider" "this" {
  client_id_list  = ["sts.amazonaws.com"]
  thumbprint_list = [data.tls_certificate.this.certificates[0].sha1_fingerprint]
  url             = aws_eks_cluster.this.identity[0].oidc[0].issuer
}
```

to your state and run `terraform plan`

High

### Logs

_No response_

### Additional Information

Works with downgrading `hashicorp/tls` provider version to `3.4.0`.

This use-case is also shown as an example in https://registry.terraform.io/providers/hashicorp/tls/latest/docs/data-sources/certificate

### Code of Conduct

detro:
Closes #244

`Blocks`, on a protocol-level, don't support expressing the attribute of being `Computed`: `terraform-plugin-framework` doesn't support it ([yet?](https://github.com/hashicorp/terraform-plugin-framework/issues/214)).

But in our case it's possible to just do away with `Blocks` for the `tls_certificate.certificates`, and instead define an attribute of type `ListType`, that carries an `ObjectType`. By using an actual `Attribute`, we can then express the fact that this is a `Computed` attribute of the data source, and has such it's not expected to have a value until the `APPLY-PHASE` of Terraform. That will make the error encountered by practitioners in #224 go away.

NOTE: We CANNOT use `Attribute.NestAttributes` for this, as that would break the compatibility with Protocol 5, i.e with `TF >= 0.12`.

### DIFF
--- a/internal/provider/data_source_certificate.go
+++ b/internal/provider/data_source_certificate.go
@@ -12,7 +12,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/schemavalidator"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
@@ -90,7 +89,7 @@ func (dst *certificateDataSourceType) GetSchema(_ context.Context) (tfsdk.Schema
 						AttrTypes: x509CertObjectAttrTypes(),
 					},
 				},
-				Computed: true,
+				Computed:            true,
 				MarkdownDescription: "The certificates protecting the site, with the root of the chain first.",
 			},
 		},
--- a/internal/provider/data_source_certificate.go
+++ b/internal/provider/data_source_certificate.go
@@ -84,82 +84,15 @@ func (dst *certificateDataSourceType) GetSchema(_ context.Context) (tfsdk.Schema
 				Computed:            true,
 				MarkdownDescription: "Unique identifier of this data source: hashing of the certificates in the chain.",
 			},
-		},
-		Blocks: map[string]tfsdk.Block{
 			"certificates": {
-				NestingMode: tfsdk.BlockNestingModeList,
-				MinItems:    0,
-				// TODO Remove the validators below, once a fix for https://github.com/hashicorp/terraform-plugin-framework/issues/421 ships
-				Validators: []tfsdk.AttributeValidator{
-					listvalidator.SizeAtLeast(0),
+				Type: types.ListType{
+					ElemType: types.ObjectType{
+						AttrTypes: x509CertObjectAttrTypes(),
+					},
 				},
-				Attributes: map[string]tfsdk.Attribute{
-					"signature_algorithm": {
-						Type:                types.StringType,
-						Computed:            true,
-						MarkdownDescription: "The algorithm used to sign the certificate.",
-					},
-					"public_key_algorithm": {
-						Type:                types.StringType,
-						Computed:            true,
-						MarkdownDescription: "The key algorithm used to create the certificate.",
-					},
-					"serial_number": {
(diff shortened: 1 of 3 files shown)


internal/provider/data_source_certificate.go
```diff
@@ -84,83 +84,16 @@ func (dst *certificateDataSourceType) GetSchema(_ context.Context) (tfsdk.Schema
...
-							"[`trimspace()`](https://www.terraform.io/language/functions/trimspace).",
+				Type: types.ListType{
+					ElemType: types.ObjectType{
+						AttrTypes: x509CertObjectAttrTypes(),
 					},
 				},
+				Computed: true,
+				Validators: []tfsdk.AttributeValidator{
+					listvalidator.SizeAtLeast(1),
+				},
```
bflad, line 96: Nit: Attribute validation only applies to practitioner configuration and the `Validate*Config` RPCs, or put differently, this has no effect here since the attribute is not `Optional` or `Required` to make it configurable

```suggestion
```

internal/provider/data_source_certificate_test.go
```diff
@@ -639,3 +639,45 @@ func TestDataSourceCertificate_MalformedURL(t *testing.T) {
...
+		Steps: []r.TestStep{
+			{
+
+				Config: `
+					resource "tls_private_key" "test" {
+						algorithm = "ED25519"
+					}
+
+					data "tls_certificate" "test" {
+						url = replace(tls_private_key.test.id, "/^.*$/","https://terraform.io")
```
bflad, line 656: Nit: May be good to leave a code comment here explaining what it happening for future readers

```suggestion
						# This attribute value must be unknown to trigger
						# the behavior, therefore this replaces the unknown
						# value with a known value on apply, so the URL is valid.
						url = replace(tls_private_key.test.id, "/^.*$/","https://terraform.io")
```

internal/provider/data_source_certificate_test.go
```diff
@@ -639,3 +639,45 @@ func TestDataSourceCertificate_MalformedURL(t *testing.T) {
...
+					}
+
+					output "test" {
+						value = data.tls_certificate.test.certificates[0].sha1_fingerprint
```
bflad, line 660: Nit: Similarly here, for future readers

```suggestion
						# This test must reference an underlying value of the
						# certificates attribute to trigger the behavior.
						value = data.tls_certificate.test.certificates[0].sha1_fingerprint
```

bflad (approved):
Looks good to me, just left some suggestions.

PR #248, opened 2022-07-25
protocol name update

vravind1:


bflad (approved):
Looks good to me, although we should probably also fix the link above this table as well.

detro:
Oh sorry @vravind1 - yesterday evening I put this up https://github.com/hashicorp/terraform-provider-tls/pull/251 and I didn't realised you were also tweaking the table.

PR #269, opened 2022-09-03
make referencing path to example generic

l-with:
since terraform provider tls is mentioned as inspiring example for [terraform-plugin-docs](https://github.com/hashicorp/terraform-plugin-docs) the reference to the examples in the documentation templates is coded generic to avoid "copy-paste-forget-to-edit" failures


l-with:
FYI: the equivalent PR for terraform-provider-random was approved and merged: https://github.com/hashicorp/terraform-provider-random/pull/309

PR #314, opened 2023-01-25
ci: merging `golangci-lint` rules + fixes

austinvalle:
contributes to:
- https://github.com/hashicorp/terraform-providers-devex-internal/issues/102

### Notes
- Fixed a couple type assertions by switching around the `Elements` looping. Not sure if this is the recommended approach? Can adjust if needed
- Skipping `paralleltest` for the providers

### DIFF
--- a/internal/provider/resource_cert_request.go
+++ b/internal/provider/resource_cert_request.go
@@ -254,8 +254,13 @@ func (r *certRequestResource) Create(ctx context.Context, req resource.CreateReq
 			"ipAddresses": newState.IPAddresses,
 		})
 
-		for _, ipElem := range newState.IPAddresses.Elements() {
-			ipStr := ipElem.(types.String).ValueString()
+		ipAddresses := make([]string, 0)
+		res.Diagnostics.Append(newState.IPAddresses.ElementsAs(ctx, &ipAddresses, false)...)
+		if res.Diagnostics.HasError() {
+			return
+		}
+
+		for _, ipStr := range ipAddresses {
 			ip := net.ParseIP(ipStr)
 			if ip == nil {
 				res.Diagnostics.AddError("Invalid IP address", fmt.Sprintf("Failed to parse %#v", ipStr))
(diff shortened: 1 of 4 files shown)


bendbennett (approved):
LGTM 👍 

The [test coverage](https://github.com/hashicorp/terraform-plugin-framework/blob/main/types/basetypes/list_test.go#L450) for using `[]string` indicates that using `ElementsAs()` in this way should work fine and I can't see any reason to use `[][]basetypes.StringValue` but I'd be interested in hearing Brian's opinion.

internal/provider/resource_cert_request.go
```diff
@@ -254,8 +254,13 @@ func (r *certRequestResource) Create(ctx context.Context, req resource.CreateReq
 			"ipAddresses": newState.IPAddresses,
 		})
 
-		for _, ipElem := range newState.IPAddresses.Elements() {
-			ipStr := ipElem.(types.String).ValueString()
+		ipAddresses := make([]string, 0)
```
bflad, line 257: Nit: If we're going to instantiate the slice with a size/capacity, we should probably do it fully since we know the information upfront:

```suggestion
		ipAddresses := make([]string, 0, len(newState.IPAddresses.Elements()))
```

Although most code examples for As/ElementsAs typically instantiate via:

```go
var ipAddresses []string
```

bflad (commented):
`ElementsAs` is preferable for this use case and will allow the framework to return an error diagnostic if the given type is not compatible with the target variable.

As for whether to use `[]types.String` or `[]string`, the latter can only accept non-null known values (not null or unknown). For `Optional` attributes during the `ApplyResourceChange` RPC (e.g. `Create` method), all values should be known (null or with a value). If we wanted to be extra cautious to prevent null values before practitioners receive a potentially confusing error here while leaving this with `[]string`, we could add a validator that requires all values to be non-null, e.g.

```go
listvalidator.ValueStringsAre(
  stringvalidator.SizeAtLeast(1),
),
```

Or similar.

internal/provider/resource_cert_request.go
austinvalle: Good call, switched them all to nil slices 👍🏻

internal/provider/resource_cert_request.go
```diff
@@ -64,6 +65,11 @@ func (r *certRequestResource) Schema(_ context.Context, req resource.SchemaReque
 				PlanModifiers: []planmodifier.List{
 					listplanmodifier.RequiresReplace(),
 				},
+				Validators: []validator.List{
```
austinvalle, line 68: @bflad added the validators

PR #330, opened 2023-03-03
Updating to Go 1.19

bendbennett:
Closes: #329


bflad (approved):
Looks good to me 🚀 Note that the README.md still says 1.17, so would be good to fix that. 👍

Issue #17, opened 2018-04-01
Passwords for certificates and keys


jackivanov:
Hi there,

I guess many people would like to set passwords to the certificates and keys. This is a base feature.
Is there any particular reason why it's not implemented yet?

Thanks!

SpencerBrown:
Passwords are used to encrypt TLS private keys, because private keys are security sensitive. TLS certificates are public, and are provided by a TLS peer during the TLS handshake protocol for validation. TLS certificates do not need keys.

For the TLS provider, we would need a password argument for:

* tls_public_key data source (to decrypt the provided private key)
* tls_private_key resource (to encrypt the generated private key)
* tls_self_signed_cert (to decrypt the provided private key)
* tls_locally_signed_cert (to decrypt the provided private key)
* tls_cert_request (to decrypt the provided private key)

The ability to [secure sensitive data](https://www.terraform.io/docs/state/sensitive-data.html) (such as passwords or unencrypted private keys) is planned for future releases of Terraform. In the meantime, the state (local or remote) needs to be secured by means external to Terraform itself.

As to why it's not implemented yet, the Terraform TLS Provider is a [community provider](https://github.com/hashicorp/terraform/blob/master/.github/CONTRIBUTING.md#hashicorp-vs-community-providers) and depends on the community to provide bug fixes and new features. For example, I am a contributor out of my own interest and on my own time. Consider becoming a contributor!

jackivanov:
Thanks! In addition we need to consider not only destroying certificates, but creating a CRL as well.

SpencerBrown:
1. Not sure what you mean by "destroying certificates". Can you amplify on this?
2. Supporting Certificate Revocation Lists (CRLs) is certainly possible but is a different topic from passwords. Do you have a specific need for CRLs? If so please open a separate issue, and help me understand the priority of this for you.

jackivanov:
Moved the CRL question to #20

rabidscorpio:
So this provider is used for much more than just TLS keys and certificates. Lots of people use this for generating SSH keys, for example, as noted by the fact the `tls_private_key` has a `public_key_openssh` attribute. I think the response given by @SpencerBrown is a little short sighted here, although that was 7(!) years ago.

[...]

PR #336, opened 2023-03-13
Adds an `encrypted_pem` resource

cimnine:
This solves one half of #17: Creating an encrypted private key.

I built this resource because encrypted private keys are required for [`container_app_environment_certificate`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/container_app_environment_certificate).

Decrypting should be very similar to do, but I currently have no need for it.


cimnine:
FYI: I closed the PR because it didn't work with passwords that contained symbols. (That could simply be added as a constraint for the password.) But for me, the main reasons to stop work on it was that (a) `azurerm_container_app_environment_certificate` didn't like the format that this resource created and (b) it didn't need an encrypted private key after all.

PR #340, opened 2023-03-19
docs: Remove read only parameter key_algorithm

torvitas:


sestegra:
@torvitas The removal should be done in `examples/provider/provider.tf`
See my commit here
https://github.com/reyerstudio/terraform-provider-tls/commit/9c194ce2196cdc209ce06a3f57e20924ec599e9e

Then, to remove the parameter in the documentation, launch `make generate` to update the documentation.

Issue #352, opened 2023-04-21
Read multiple PEM encoded certificates when using `content` in a data source


ropnop:
### Terraform CLI and Provider Versions

```
Terraform v1.3.8
on darwin_amd64
+ provider registry.terraform.io/hashicorp/tls v4.0.4
```

### Use Cases or Problem Statement

Allow parsing multiple PEM encoded certificates when using the `content` attribute in a `data` block.

I would like to be able to use read and store multiple certificates from a single PEM file that contains a CA chain. Given a file like this (cachain.pem)

```
-----BEGIN CERTIFICATE-----
MII...
...etc..
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
MII...
...etc..
-----END CERTIFICATE-----
```

I would like to be able to load it via a data resource and reference each cert enclosed within it like so:

```
data "tls_certificate" "example_content" {
  content = file("cachain.pem")
}

output "num_certs" {
  value = length(data.tls_certificate.example_content.certificates)
}
```

Currently `num_certs` is 1 because the data source stops reading after the first PEM certificate in the file.

### Proposal

I would like to update the [Read](https://github.com/hashicorp/terraform-provider-tls/blob/main/internal/provider/data_source_certificate.go#L147-L153) function when using the `content` attribute to loop through the content and append each PEM block, instead of only reading the first valid PEM block in the content. 

The resource type already stores `certificates` as a List type, so no other changes should be necessary. I also think this won't affect any backwards compatibility, since anyone currently using the `content` attribute is already referencing the data as a list with a 0 index.

### How much impact is this issue causing?

Low

### Additional Information

Will open a PR for this, but wanted to start an issue for the discussion/thoughts beforehand.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #356, opened 2023-05-10
Parse multiple certificates from single data source

ropnop:
Hello! This addresses the enhancement I raised in #352 

This moves the decoding into a loop, which breaks when there are no more PEM blocks in the provided content. 

I also added a test using the same test data CA chain.

Let me know if you think this is a valid approach. I think this should be fine to be backwards compatible since the data source returned a list (of always length 1) anyway


clintmiller:
Any update on this? Sure would be great to have this merged.

LudovicTOURMAN:
I'm also looking for this to be fixed :+1:

PR #395, opened 2023-08-03
Add support for openssh comments when generating a private key

### LINKED ISSUE #227: Allow specifying OpenSSH Private Key Comment

master

`openssh.MarshalPrivateKey` is always with `""` comment. We actually need to be able to configure it.

Here it is:

https://github.com/hashicorp/terraform-provider-tls/blob/9781d20b56443ebeaf985f2c9300fa1dd5ea94f1/internal/provider/resource_private_key.go#L226

### Proposal

Add an input variable and make it configurable. A good name would be `openssh_comment`. It can be optional, and default to an empty string for backward compatibility.

High

### Additional Information

Having a comment in the private key specifically is required for us unfortunately, and without this we can't use this provider.

We need it to make the right comment to appear when the key is added to the `ssh-agent` because we need to be able to tell what key is loaded.

### Code of Conduct

F21:
This PR adds the `openssh_comment` attribute to allow the addition of a comment to openssh keys.

Closes #227


aalvarezaph:
Any news? Would be useful for scenarios like https://github.com/hashicorp/terraform-provider-tls/issues/227#issuecomment-1232798060

srgustafson8:
Agreed - exactly the same use case I have. Can @bendbennett @bflad help (as the most prolific contributors)?

socratesx-asimov:
Come on guys please, this is what we also need.

stephybun (changes requested):
Thanks for this PR @F21! Overall this looks great, there are two minor comments which would be good to have addressed. Once that's done and the PR is rebased I can run the CI on this and we should be good to go!

Issue #360, opened 2023-05-21
Support for KeyUsage in CSR


ChevronTango:
### Terraform CLI and Provider Versions

terraform 1.4.6
tls provider 4.0.4

### Use Cases or Problem Statement

Currently the CertificateSigningRequest generated by tls_cert_request has a limited number of customizable fields. My current use case has me needing to specify both the `keyUsage` and the `extendedKeyUsage` of the certificate, and the only way I can do that right now is to use openssl to generate the CSR rather than terraform.

### Proposal

add the `keyUsage` and the `extendedKeyUsage` fields to tls_cert_request

### How much impact is this issue causing?

Medium

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ChevronTango:
I think it would be worth considering supporting all of the [Standard Extensions](https://www.openssl.org/docs/manmaster/man5/x509v3_config.html)

PR #430, opened 2023-10-28
Adding ext key usages support for csr

preved911:
Hello. We use this provider `tls_cert_request` resource to create csr, which will be used to issue certificates over certificate authority. We need to have possibility to specify some ext key usages for certificate, which we want to get. This changes intended to solve our needs. It should (maybe partially) support functionality, which was requested in https://github.com/hashicorp/terraform-provider-tls/issues/360


simonostendorf:
Would like to use this also.

simonostendorf:
Hey, since nothing has happened here since 2023, I forked the provider and added `allow_uses` and `ca_constraint` based on the PR here and my own code.

If you are interested in the changes you can find them here: https://github.com/simplifyk8s/terraform-provider-tls

Issue #57, opened 2019-09-30
Specify signature algorithm for certificate creation


seboudry:
In resource `tls_self_signed_cert` it'll be helpful to define `SignatureAlgorithm`.
For example, currently Spring Security OAuth only support `sha1WithRSAEncryption`...

PeterBurner:
Yes, this is much needed!
AWS Code Signing Profile for CC3220 also requires SHA-1 due to hardware limitations

PR #512, opened 2024-05-10
Adding support for specifying hashing algorithm

ChevronTango:
This PR adds the ability to specify the hashing algorithm that the certificate should use in its signatures.

Rather than specifying the full signature algorithm, we instead specify only the hashing algorithm as we can infer the remainder from the type of certificate supplied. This allows us the flexibility to customize the output without introducing too much complexity for the end user to manage.

This addresses https://github.com/hashicorp/terraform-provider-tls/issues/57


ChevronTango:
@austinvalle would you be free to review this and some of the other community submitted PRs? Many thanks

PR #517, opened 2024-05-17
datasource/tls_certificate: Use local TLS server and certificate for verify failure testing

### LINKED ISSUE #516: datasource/tls_certificate: TestAccDataSourceCertificate_BadSSL test failures

N/A and `main`

### Terraform Configuration

```terraform
data "tls_certificate" "test" {
  url          = "https://untrusted-root.badssl.com/"
  verify_chain = false
}
```

### Expected Behavior

Acceptance test passes without modification over time.

### Actual Behavior

```
=== RUN   TestAccDataSourceCertificate_BadSSL
    data_source_certificate_test.go:166: Step 2/2 error: Check failed: Check 14/14 error: data.tls_certificate.test: Attribute 'certificates.1.sha1_fingerprint' expected "6922cd864f3c6299f6e751a019e5ddcdbc415a71", got "eede8b066561000952c3e599d4873eed75512a3b"
--- FAIL: TestAccDataSourceCertificate_BadSSL (0.64s)
```

### Steps to Reproduce

1. `terraform apply`

Low

### Logs

_No response_

### Additional Information

The acceptance test already notes that this value is expected to change over time (presumably when the external SSL certificate is rotated).

### Code of Conduct

bflad:
Closes #516

Previously (likely due to external SSL certificate rotation):

```
=== RUN   TestAccDataSourceCertificate_BadSSL
    data_source_certificate_test.go:166: Step 2/2 error: Check failed: Check 14/14 error: data.tls_certificate.test: Attribute 'certificates.1.sha1_fingerprint' expected "6922cd864f3c6299f6e751a019e5ddcdbc415a71", got "eede8b066561000952c3e599d4873eed75512a3b"
--- FAIL: TestAccDataSourceCertificate_BadSSL (0.64s)
```

The goal of this test is to ensure the data source returns an error if there is an invalid SSL certificate chain, which can be accomplished by running a local TLS server with expired or otherwise invalid SSL certificate.

There still is one external, real-world URL test with `TestAccDataSourceCertificate_TerraformIO`. It seems important to ensure there is one valid URL test for complete coverage though. If that test becomes a regular problem, a local TLS server could potentially be spun up with a valid SSL certificate via Let's Encrypt or something, however that effort is not being prioritized at the moment.

### DIFF
--- a/internal/provider/data_source_certificate_test.go
+++ b/internal/provider/data_source_certificate_test.go
@@ -159,9 +159,6 @@ func TestAccDataSourceCertificate_TerraformIO(t *testing.T) {
 	})
 }
 
-// NOTE: Yes, this test is fetching a live certificate.
-// It can potentially break over time, and we will need to keep the
-// data we check against up to date, when that happens.
 func TestAccDataSourceCertificate_BadSSL(t *testing.T) {
 	server, err := tu.NewHTTPServer()
 	if err != nil {
--- a/internal/provider/data_source_certificate_test.go
+++ b/internal/provider/data_source_certificate_test.go
@@ -163,46 +163,33 @@ func TestAccDataSourceCertificate_TerraformIO(t *testing.T) {
 // It can potentially break over time, and we will need to keep the
 // data we check against up to date, when that happens.
 func TestAccDataSourceCertificate_BadSSL(t *testing.T) {
-	r.Test(t, r.TestCase{
+	server, err := tu.NewHTTPServer()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer server.Close()
+	go server.ServeTLS()
+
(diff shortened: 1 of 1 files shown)


internal/provider/data_source_certificate_test.go
```diff
@@ -163,46 +163,33 @@ func TestAccDataSourceCertificate_TerraformIO(t *testing.T) {
 // It can potentially break over time, and we will need to keep the
```
SBGoods, line 163: nit: Do we still need this note if we're no longer using a live certificate?

SBGoods (approved):

internal/provider/data_source_certificate_test.go
bflad: Ha, nope, thank you will remove.

Issue #35, opened 2018-11-27
Add support for `email_addresses` in `tls_cert_request` and `tls_self_signed_cert` resources


trajano:
### Terraform Version

```
Terraform v0.11.9
+ provider.aws v1.46.0
+ provider.template v1.0.0
+ provider.tls v1.2.0
```

### Affected Resource(s)

- tls_self_signed_cert 
- tls_cert_request

### Terraform Configuration Files
```hcl
resource "tls_cert_request" "client" {
  key_algorithm   = "${tls_private_key.client.algorithm}"
  private_key_pem = "${tls_private_key.client.private_key_pem}"

  subject {
    common_name = "client"
  }

  # this is what's missing
  # rfc822 = "myemail@foo.com"

  # or more easily 
  # subject {
  #  common_name = "client"
  # email_address = "foo@foo.com" 
  # this should generate the legacy support ones which is the email address as part of the subject but also add the rfc822 SAN 
  # }
}
```

### Debug Output
n/a

### Panic Output
n/a

### Expected Behavior
Should support it so we can have name and email addresses to be assigned to client certificates.

### Actual Behavior
n/a

### Steps to Reproduce
n/a

### Important Factoids

https://tools.ietf.org/html/rfc5280#section-4.1.2.6 talks about the emailAddresses

### References

#27

detro:
The `x509.Certificate` type offers a field `EmailAddresses` that sits next to `URIs`, `IPAddresses` and `DNSNames`.

This effort would be a matter of exposing the additional field and provide test coverage.

PR #405, opened 2023-08-26
#35 - Implement email Address field into Cert Request resource

gerardlemetayerc:
Hello,

I've added email address into cert request generation (as I had the need and someone had requested this in the issue #35.

```hcl
resource "tls_cert_request" "example" {
  private_key_pem = file("private_key.pem")

  subject {
    common_name  = "example.com"
    organization       = "ACME Examples, Inc"
    email_address     = "test@example.com"
  }
}
```

I've tested on my side and worked correctly. I've updated the documentation with related elements.

Have a good day!

PR #548, opened 2024-08-19
add support email address

iishabakaev:
Same as https://github.com/hashicorp/terraform-provider-tls/pull/405, because PR not merged since 2018

Fix this problem https://github.com/hashicorp/terraform-provider-tls/issues/35


iishabakaev:
@austinvalle hello!

Sorry to bother you, could you review this PR, this?
Its will be great if provider support email address field in csr generation

docs/cdktf/python/resources/cert_request.md
austinvalle, line 1: This file will get autogenerated after you merge, so you can undo the changes to `docs/cdktf`

docs/cdktf/python/resources/self_signed_cert.md
austinvalle, line 1: This file will get autogenerated after you merge, so you can undo the changes to `docs/cdktf`

internal/provider/resource_cert_request.go
```diff
@@ -193,6 +193,13 @@ func (r *certRequestResource) Schema(_ context.Context, req resource.SchemaReque
 							},
 							Description: "Distinguished name: `SERIALNUMBER`",
 						},
+						"email_address": schema.StringAttribute{
+							Optional: true,
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.RequiresReplace(),
+							},
+							Description: "Email Address: `1.2.840.113549.1.9.1`",
```
austinvalle, line 201: Just a small change to be a little more specific what the value is
```suggestion
							Description: "ASN.1 Object Identifier (OID): `1.2.840.113549.1.9.1`",
```

internal/provider/common_cert.go
```diff
@@ -386,5 +386,12 @@ func createSubjectDistinguishedNames(ctx context.Context, subject certificateSub
 		result.SerialNumber = subject.SerialNumber.ValueString()
 	}
 
+	if !subject.EmailAddress.IsNull() && !subject.EmailAddress.IsUnknown() {
+		result.ExtraNames = append(result.ExtraNames, pkix.AttributeTypeAndValue{
+			Type:  []int{1, 2, 840, 113549, 1, 9, 1},
```
austinvalle, line 391: I know the underlying type is an alias for `[]int`, but would still prefer to use the aliased type here from `encoding/asn1`:
```go
Type: asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 9, 1},
```

internal/provider/common_cert.go
```diff
@@ -386,5 +386,12 @@ func createSubjectDistinguishedNames(ctx context.Context, subject certificateSub
...
+			Value: subject.EmailAddress.ValueString(),
```
austinvalle, line 392: The email address string here will be `UTF-8` as that's the standard encoding for Terraform. [The RFC](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.6) states this should `IA5String`:
> The attribute value for emailAddress is of type
> IA5String to permit inclusion of the character '@', which is not part
> of the PrintableString character set.

I read through some online discussions and it seems something like this should be a suitable alternative (although I have not tested myself):
```go
Value: asn1.RawValue{
  Tag:   asn1.TagIA5String, 
  Bytes: []byte(subject.EmailAddress.ValueString()),
},
```

austinvalle (changes requested):
Hey @iishabakaev 👋🏻 , thanks for submitting the PR and apologies for the delay in review!

I have left some specific code feedback on the PR as well as I have some general feedback that will need to be addressed before we can accept the change:
- This PR only updates the `tls_cert_request` schema to include this new field, however the underlying logic is also shared by `tls_self_signed_cert`.
  - I think we should adjust the `tls_self_signed_cert` schema as well to include this new `email_address` field so we can fully close #35 
- This PR will need acceptance tests before we can merge it. This should just be updating the existing ones that test the subject and just verifying the data is properly set:
  - https://github.com/hashicorp/terraform-provider-tls/blob/10b2c5511703aab6de581d2e0fa6134fe64e12d6/internal/provider/resource_cert_request_test.go
  - https://github.com/hashicorp/terraform-provider-tls/blob/10b2c5511703aab6de581d2e0fa6134fe64e12d6/internal/provider/resource_self_signed_cert_test.go
- This PR will [need a changelog](https://github.com/hashicorp/terraform-provider-tls/blob/main/.github/CONTRIBUTING.md#changelog), but I can also help with that once the PR is updated to address some of the feedback

austinvalle:
Some general housekeeping things:
- You'll need to run the documentation generator to update the markdown docs: [`make generate`](https://github.com/hashicorp/terraform-provider-tls/blob/10b2c5511703aab6de581d2e0fa6134fe64e12d6/GNUmakefile#L13-L15)
- Our CI acceptance tests won't run from a forked PR, so once you make the changes and test locally, I'll open a mirror PR to confirm the tests are all green ✅

iishabakaev:
Updated PR based on all comments

For testing I had to tinker a bit, the standard function x509.ParseCertificate parses ExtraNames in raw, so I added the creation of email_address

docs/resources/cert_request.md
austinvalle, line 1: Looks like the documentation generator still needs to be run on this branch. There should be another documentation update in `docs/resources/self_signed_cert.md`

austinvalle (commented):
Thanks for the updates @iishabakaev!

I've opened a mirror PR which we can eventually use to run the CI tests (I still need to manually move the changes/approve them 😄 )
https://github.com/hashicorp/terraform-provider-tls/pull/599

Currently I think the remaining items for this PR are:
- Run the doc generator locally to fix that diff in our CI
- Remove the changes to `docs/cdktf`
- Verify the tests are all green ✅ 

After that I can push a changelog and merge it.

docs/resources/cert_request.md
iishabakaev: Done, I hope I didn't miss anything

internal/provider/resource_self_signed_cert_test.go
```diff
@@ -586,6 +596,7 @@ func selfSignedCertConfig(validity, earlyRenewal uint32) string {
                 country = "US"
                 postal_code = "95559-1227"
                 serial_number = "2"
+				email_address = "example@example.com"
```
austinvalle, line 599: You won't be able to add this here since tests like `TestAccResourceSelfSignedCert_UpgradeFromVersion3_4_0 ` also use this function, which tests migration from an older version of the provider that doesn't have the `email_address` attribute

https://github.com/hashicorp/terraform-provider-tls/actions/runs/12956793277/job/36143882711?pr=599

You can duplicate the config in `TestResourceSelfSignedCert` to just add email address there, or make it a parameter in this function, whichever you prefer

austinvalle (commented):
Looks like there is a test failure in the PR, see comment in review.

Also you can run the tests locally by just running `make testacc` or just running with any `go test` command equivalent with the env variable `TF_ACC=1`

docs/resources/cert_request.md
austinvalle: Docs look good 👍🏻

internal/provider/resource_self_signed_cert_test.go
iishabakaev: Sorry for delay

I copied dublicate config for ver 340, but cant run this test locally

Seems like version 3.4.0 dont have in registry
```
=== RUN   TestAccPrivateKeyRSA_UpgradeFromVersion3_4_0
    resource_private_key_test.go:65: TestStep 1/2 running init: exit status 1
        
        Error: Failed to query available provider packages
        
        Could not retrieve the list of available versions for provider hashicorp/tls:
        could not connect to registry.terraform.io: failed to request discovery
        document: Get "https://registry.terraform.io/.well-known/terraform.json": EOF
```

All another tests is passed

austinvalle (dismissed):
Sorry for the long delay on your PR review! 🚀 

I've merged a PR from the original contributor of the feature in https://github.com/hashicorp/terraform-provider-tls/pull/608, but there's a lot of work your did in this review that I'd also like to merge.

You'll see some commits/pushes from myself as I clean up the merge conflicts, and then I'll merge/release. Thanks for your contribution!

Issue #629, opened 2025-03-16
Introduction of Terraform ephemeral resources for `tls_private_key` and `tls_public_key`


JakubSzuber:
### Terraform CLI and Provider Versions

Starting 1.11.X

### Use Cases or Problem Statement

The `tls_private_key` resource and `tls_public_key` data source are stored in Terraform state files and plan files which is not a great idea because of security concerns.

### Proposal

Starting 1.11.X Terraform introduces ephemeral resources that allow for handling temporary values:  https://developer.hashicorp.com/terraform/language/resources/ephemeral

Introduction of ephemeral resources `tls_private_key` and `tls_public_key` would allow to temporarily get the value of private/public key and then store it e.g. as AWS SSM parameter or secret in AWS Secret Manager - it would allow for full automation and security!

We already have ephemeral resources in the AWS provider that can integrate very well with proposed `tls_private_key` and `tls_public_key` ephemeral resources:
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/ephemeral-resources/ssm_parameter
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/ephemeral-resources/secretsmanager_secret_version
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/ephemeral-resources/secretsmanager_random_password
- https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key

### How much impact is this issue causing?

High

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

pkeecom:
I second that. Please implement.

JakubSzuber:
Yeah, or any other cloud secret storage - Azure Vault, AWS SSM Parameter store, AWS SecretManger, etc.

guanwee-loo:
Hi, one use case of ephemeral tls_private_key/tls_public_key is in the creation of AWS key pairs.  In this case, the aws_key_pair will need to support write-only attributes.

hostrowskisii:
I already created an issue for this too - https://github.com/hashicorp/terraform-provider-aws/issues/41922

JakubSzuber:
Right. There is created issue for this already - https://github.com/hashicorp/terraform-provider-aws/issues/41922

austinvalle:
[...]

PR #637, opened 2025-04-01
Support `ephemeral.tls_private_key `

lonegunmanb:
This pr added a new ephemeral resource `tls_private_key` to this provider, it should solve #629 partially.


lonegunmanb:
For anyone who want to preview this feature: https://registry.terraform.io/providers/lonegunmanb/ephemeraltls/latest

austinvalle (approved):
Hey @lonegunmanb 👋🏻 ,

Looks good to me! There are a couple comments I left on the PR, however we're planning on releasing the provider this week and I'd like to include this change with that release.

I'm going to push a couple changes to your PR, mostly documentation and changelogs and then merge (it'll all get squashed into a single commit under your git email).

Thanks for the great PR 🚀

JonasNuvibit:
the tls_cert_request should also support the ephemeral private key as input. 

With the current provider it fails with the following error: 
```
Error: Invalid use of ephemeral value

Ephemeral values are not valid for "private_key_pem", because it is not a
write-only attribute and must be persisted to state.
```

austinvalle:
> the tls_cert_request should also support the ephemeral private key as input.
> 
> With the current provider it fails with the following error:
> 
> ```
> Error: Invalid use of ephemeral value
> 
> Ephemeral values are not valid for "private_key_pem", because it is not a
> write-only attribute and must be persisted to state.
> ```

Hey @JonasNuvibit 👋🏻 , that is correct, at the moment none of the TLS managed resources have a write-only attribute, which seems like a reasonable request.

That falls outside the scope of this PR/original issue, so if you're looking for a new write-only attribute in `tls_cert_request` or any of the other managed resources you can [create a new issue](https://github.com/hashicorp/terraform-provider-tls/issues/new?template=Feature_Request.yml) for folks to vote/comment on 👍🏻

bbasata (approved):

PR #646, opened 2025-04-23
chore: Final onboarding to CRT

austinvalle:
Now that the provider is fully onboarded to CRT and has a successful release, this PR removes the prior release workflow/goreleaser config.


bbasata (approved):
:shipit:

Issue #645, opened 2025-04-22
Write-only attribute support for TLS managed resources


JonasNuvibit:
### Terraform CLI and Provider Versions

terraform >= 1.10
tls provider: v4.1.0 (to be relased) https://github.com/hashicorp/terraform-provider-tls/milestone/8
Tested with open PR: #637  https://github.com/hashicorp/terraform-provider-tls/pull/637  
https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/self_signed_cert

### Use Cases or Problem Statement

With ephemeral resources comming to the provider, write-only attribute support should be added to the existing resources. 

While testing the [ephemeral.tls_private_key](https://github.com/hashicorp/terraform-provider-tls/pull/637) feature I noticed that the `tls_cert_request` does not support the ephemeral key as `private_key_pem` attribute.

This impacts the following resources:
- tls_cert_request
- tls_locally_signed_cert
- tls_self_signed_cert

### Proposal

Write-only attribute support (private_key_pem) for the following resources:
- tls_cert_request
- tls_locally_signed_cert
- tls_self_signed_cert

### How much impact is this issue causing?

High

### Additional Information

# To reproduce

```
# The private key is generated using ephemeraltls provider
ephemeral "ephemeraltls_private_key" "key" {
  algorithm = "RSA"
  rsa_bits  = 4096
}

resource "tls_cert_request" "csr" {
  private_key_pem = ephemeral.ephemeraltls_private_key.key.private_key_pem
  subject {
    common_name = "foo-bar"
  }
}
```

Error message:
```
Error: Invalid use of ephemeral value

Ephemeral values are not valid for "private_key_pem", because it is not a
write-only attribute and must be persisted to state.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

sanderjochems-capgemini:
@lonegunmanb I see you made a lot of work on the ephemeral resources for this provider. Thank you for that! Would you be able to take a look at this issue and maybe even create a PR, please?

tobywan:
I have found that I cannot use an ephemeral azure keyvault secret to provide the `private_key_pem` of a `data.tls_public_key` resource,

e.g. in file `secrets.tf`

```
ephemeral "azurerm_key_vault_secret" "example" {
  name         = "existing_secret"
  key_vault_id = azurerm_key_vault.example.id
}

data "tls_public_key" "example" {
[...]

PR #689, opened 2025-08-16
feat: support ephemeral private keys for self signed certificates

mvanholsteijn:
Adds the properties private_key_pem_wo and private_key_pem_wo_version to support the use of ephemeral private keys for self signed certificates, keeping the private key out of the state file.

## Related Issue

#645

## Description

Added support for a write-only private_key_pem as described in  https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments

## Rollback Plan

nope.


josh-abram:
How is this not merged yet! Has Hashicorp/IBM abandoned this provider?

PR #692, opened 2025-08-21
Remove Read-Only Attribute in provider example doc

jinjiaKarl:
## Related Issue

Remove key_algorithm (Read-Only Attribute) in provider example doc

## Description

When running example provided in the provider example, it has the error.

```
Error: Invalid Configuration for Read-Only Attribute
│
│   with tls_self_signed_cert.example,
│   on main.tf line 15, in resource "tls_self_signed_cert" "example":
│   15:   key_algorithm = tls_private_key.example.algorithm
│
│ Cannot set value for this attribute as the provider has marked it as read-only. Remove the configuration line setting the value.
│
│ Refer to the provider documentation or contact the provider developers for additional information about configurable and read-only attributes that are
│ supported.
```

## Rollback Plan

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -37,7 +37,6 @@ resource "tls_private_key" "example" {
 }
 
 resource "tls_self_signed_cert" "example" {
-  key_algorithm   = tls_private_key.example.algorithm
   private_key_pem = tls_private_key.example.private_key_pem
 
   # Certificate expires after 12 hours.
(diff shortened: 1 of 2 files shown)


austinvalle (approved):
Thanks for the fix @jinjiaKarl ! Looks like this was missed in https://github.com/hashicorp/terraform-provider-tls/issues/174 and https://github.com/hashicorp/terraform-provider-tls/pull/215 ---> 🚀

PR #695, opened 2025-08-30
adds support for write only private key on tls_cert_request

mvanholsteijn:
## Related Issue

Contributes to https://github.com/hashicorp/terraform-provider-tls/issues/645

## Description

Adds support for write only private key on tls_cert_request as described in https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments

Adds the write only properties as an alternative for private_key_pem.

- private_key_pem_wo
- private_key_pem_wo_version

## Rollback Plan


ghost:
@austinvalle and @bbasata can you have a look at this PR, please? I see you two reviewed the other write only PRs too

PR #764, opened 2026-04-13
enhancement: force HTTP/2 when fetching peer certificates via HTTPS

daper:
## Related Issue

Fixes # 

## Description

Added `ForceAttemptHTTP2: true` to the `http.Transport` in the `fetchPeerCertificatesViaHTTPS` function. This ensures that when the `tls_certificate` data source fetches peer certificates over HTTPS, it will prefer HTTP/2 connections where the server supports it, improving connection efficiency.

The change is minimal and targeted: it sets the `ForceAttemptHTTP2` field on the existing transport struct alongside the existing `TLSClientConfig` and `Proxy` fields.

> **Note:** A changelog entry via Changie should be added once a PR/issue number is assigned:
> ```
> data-source/tls_certificate: Force HTTP/2 connection attempts when fetching certificates via HTTPS
> ```

## Rollback Plan

No changes to security controls. `ForceAttemptHTTP2` only affects the HTTP protocol version negotiated; the existing `TLSClientConfig` (including `InsecureSkipVerify`) and proxy settings are unchanged.

### DIFF
--- a/internal/provider/data_source_certificate.go
+++ b/internal/provider/data_source_certificate.go
@@ -237,6 +237,7 @@ func fetchPeerCertificatesViaHTTPS(ctx context.Context, targetURL *url.URL, shou
 
 	client := &http.Client{
 		Transport: &http.Transport{
+			ForceAttemptHTTP2: true,
 			TLSClientConfig: &tls.Config{
 				InsecureSkipVerify: !shouldVerifyChain,
 			},


ansgarm:
Hi @daper 👋
Thank you for the PR – out of curiosity what is the underlying use-case or issue you were running into that made you raise this PR? Did you run into any problems with the current implementation?

daper:
> Hi @daper 👋 Thank you for the PR – out of curiosity what is the underlying use-case or issue you were running into that made you raise this PR? Did you run into any problems with the current implementation?

Hi, yes, sorry I should have given more context. I tried to use it against a web server that did not have support for http/1.1 and it failed the terraform execution. By adding this, it tries with http/2 falling back to 1 if unsupported.

daper:
Hi there, can I do anything to speed up the review process?

PR #793, opened 2026-07-16
tests: update DevDot cert chain assertions to match current chain

rainkwan:
## Related Issue

Fixes CI errors for PR #789 

## Description
The `TestAccDataSourceCertificate_DevDot` test fetches the live TLS cert chain for `developer.hashicorp.com` and asserts against hardcoded values. The chain has rotated — it now returns 3 certs instead of 2, with a new ISRG root (`CN=Root YR`) and a new Let's Encrypt intermediate (`CN=YR2`).

This updates the assertions to match the current chain. As noted in the existing comment, this is expected maintenance.

Changes:
- `certificates.#` updated from `2` to `3`
- `certificates.0` now describes `CN=Root YR` (ISRG root)
- `certificates.1` now describes `CN=YR2` (Let's Encrypt intermediate)
- `certificates.2` now describes `CN=developer.hashicorp.com` (leaf)
- Fingerprints are not asserted on the intermediate or leaf (both rotate frequently)

## Rollback Plan

- [x] If a change needs to be reverted, we will roll out an update to the code within 7 days.

No

### DIFF
--- a/internal/provider/data_source_certificate_test.go
+++ b/internal/provider/data_source_certificate_test.go
@@ -123,7 +123,7 @@ func TestAccDataSourceCertificate_UpgradeFromVersion3_4_0(t *testing.T) {
 // NOTE: Yes, this test is fetching a live certificate.
 // It will break over time and we will need to keep the
 // data we check against up to date, when that happens.
-// Last updated: 2025-07-02
+// Last updated: 2025-07-02.
 func TestAccDataSourceCertificate_DevDot(t *testing.T) {
 	r.Test(t, r.TestCase{
 		ProtoV5ProviderFactories: protoV5ProviderFactories(),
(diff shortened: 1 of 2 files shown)


internal/provider/data_source_certificate_test.go
```diff
@@ -123,6 +123,7 @@ func TestAccDataSourceCertificate_UpgradeFromVersion3_4_0(t *testing.T) {
 // NOTE: Yes, this test is fetching a live certificate.
 // It will break over time and we will need to keep the
 // data we check against up to date, when that happens.
+// Last updated: 2025-07-02.
```
michele-berg, line 126: Is this date a typo?

internal/provider/data_source_certificate_test.go
```diff
@@ -135,26 +136,32 @@ func TestAccDataSourceCertificate_DevDot(t *testing.T) {
 					}
 				`,
 				Check: r.ComposeAggregateTestCheckFunc(
-					r.TestCheckResourceAttr("data.tls_certificate.test", "certificates.#", "2"),
+					r.TestCheckResourceAttr("data.tls_certificate.test", "certificates.#", "3"),
```
michele-berg, line 139: Updating the count to match the current chain makes sense. One thing I’m wondering, though - are we still asserting more of the live chain structure here than we need for this test? Since the endpoint’s intermediates/roots can rotate independently of provider behavior, tightening the assertions to the minimum invariants we care about could help avoid future maintenance-only breakages.

michele-berg (approved):

PR #807, opened 2026-08-31
supress vuln scan for unused deprecated dependency

ansgarm:
golang.org/x/crypto/openpgp is deprecated/unmaintained with no fixed
version. The provider does not use this package (confirmed via
`go mod why golang.org/x/crypto/openpgp`); it imports the maintained
fork github.com/ProtonMail/go-crypto/openpgp instead.


rainkwan (approved):
LGTM

